PR #16, opened 2022-01-28
US132939: add system access card

johngwilkinson:
This just uses stub data to lay out the card. It's quite likely it will need tweaking but I'd like to get the basics in to support development of other cards. Note vis diffs only for testing so far: unit tests for correct data don't make sense until we clarify the server hook-up.

Quality Notes

- functional
  - demo page
- security/devops/ux/docs: n/a

FYI @Brightspace/draco-vis 

image


components/cards/system-access.js
```diff
@@ -0,0 +1,255 @@
...
+						color: 'var(--d2l-color-ferrite)',
+						fontFamily: 'Lato'
+					}
+				},
+
+				// Highcharts picks UTC times, or optionally browser-local times. But we want to match the server timezone, so
+				// we select some or all of the dates in the actual data
+				tickPositioner: function() {
+					return !that._data || that._data.length === 0 ? undefined :
+						that._data.map(v => v[0]).filter((v, i) => i % Math.ceil(that._data.length / 10) === 0);
```
bemailloux, line 158: It's weird to me that ticks aren't placed at even intervals (at least according to the screenshots). Maybe they should be?

components/chart/chart.js
```diff
@@ -0,0 +1,205 @@
...
+	'<div>{xAxisDescription}</div>' +
+	'<div>{yAxisDescription}</div>' +
+	'<div>{annotationsTitle}{annotationsList}</div>';
+
+/**
+ * based on highcharts-webcomponent (npm) - main modifications: convert to plain .js, fix import, fix typing,
+ * remove flag for suppressing updates, prevent unneeded update on first render
+ * @element highcharts-chart
+ */
+class Chart extends SkeletonMixin(Localizer(LitElement)) {
```
bemailloux, line 34: Is this what's being used in AQ and Engagement? Or is it something new?

components/cards/system-access.js
johngwilkinson: There's a tick on each data point, but my sample data isn't evenly distributed. So this demonstrates that doing the ticks this way works in general, but it will look even when we get server data, which should have an entry for every date in the range.

components/chart/chart.js
johngwilkinson: Yeah, same thing. I'll make a story for pulling this out into something reusable.

bemailloux (approved):

components/cards/system-access.js
```diff
@@ -0,0 +1,255 @@
...
+import { formatDate } from '@brightspace-ui/intl/lib/dateTime';
+import { Localizer } from '../../locales/localizer';
+import { MobxLitElement } from '@adobe/lit-mobx';
+import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
+
+class SystemAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
+
+	static get properties() {
+		return {
+			isDemo: { type: Boolean, attribute: 'demo' }
```
Vitalii-Misechko, line 15: `thought` 
We have managed to get rid of `isDemo` - flag in AQD project. Should we keep that practice in this project too?

components/chart/chart.js
```diff
@@ -0,0 +1,205 @@
...
+
+	disconnectedCallback() {
+		if (this.chart) {
+			this.chart.destroy();
+			this.chart = null;
+		}
+	}
+
+	get chartContainer() {
+		return this.shadowRoot.querySelector('#chart-container');
```
Vitalii-Misechko, line 185: We need a new point in a quality plan `Lit 2 Phase 1 compliant` [until all projects migrated to Lit 2](https://d2l.slack.com/archives/C02TNRJT64X/p1643655077398579)

Vitalii-Misechko: violated rule
`add non-null checks to shadowRoot accesses (not including test files)`

components/overlay.js
```diff
@@ -0,0 +1,58 @@
...
+			:host([skeleton]) {
+				height: 100%;
+				width: 100%;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+			.d2l-insights-overlay-container {
+				align-items: center;
+				background-color: var(--d2l-color-white);
```
Vitalii-Misechko, line 35: There is no `--d2l-color-white` anymore.  Could you use `white` instead

Vitalii-Misechko (commented):
I've added a few suggestions. Also, I see a skeleton code but I don't see any visual test with it. Is it coming in some future PR?

components/chart/chart.js
johngwilkinson: What is checking that rule for you? Is this something I could add here?

components/cards/system-access.js
johngwilkinson: Yes, removing it in the next PR

components/overlay.js
johngwilkinson: done, thanks

components/chart/chart.js
johngwilkinson: copied the updated chart.js from AQD

Vitalii-Misechko (approved):
LGTM!

PR #20, opened 2022-02-01
US132939: load server data

johngwilkinson:
I've matched the current server response for now, but that may change slightly. Hoping to get this foundational stuff in in the meantime.

Todo in another PR: error handling (currently sets a flag but doesn't use it)

Quality Notes
- functional
  - demo & e2e (server data is currently canned): graph appears as expected
- security/performance/devops/ux/docs: n/a
- lit2: no new shadowRoot references, I believe

FYI @Brightspace/draco-vis


model/dataApiClient.js
```diff
@@ -0,0 +1,79 @@
...
+ * Wraps the BDP API adoption data routes and related LMS calls
+ */
+export class DataApiClient {
+	constructor(metronEndpoint) {
+		this._metronEndpoint = metronEndpoint;
+		this._locations = null;
+		this._launchRequestCompleted = null;
+	}
+
+	startDataRequest({ orgUnitIds, userSettings }) {
```
bemailloux, line 24: should this be async?

bemailloux (commented):
Haven't looked in too much detail but makes sense to me. I like how it mirrors the back-end.

model/dataApiClient.js
johngwilkinson: Idea is that this creates a promise so all call to fetchResultSet can then await on that same promise. This method itself can return right away though.

johngwilkinson:
I think doing most testing via vis diffs is probably best for simplicity and completeness, but if there's anything that seems to call for unit tests here, please call it out.

model/data.js
```diff
@@ -0,0 +1,17 @@
...
+
+export class Data {
+	constructor({ resultSetNames, dataApiClient }) {
+		this._resultSets = resultSetNames.map(name => new ResultSet({ name, dataApiClient }));
+		this._dataApiClient = dataApiClient;
+	}
+
+	load({ orgUnitIds, userSettings }) {
+		this._dataApiClient.startDataRequest({ orgUnitIds, userSettings });
+		this._resultSets.forEach(resultSet => resultSet.load());
```
Vitalii-Misechko, line 11: Should it return an array of promises so a caller can wait on results? It seems we're not going to wait on results but let MobX do the trick

route-loader.js
```diff
@@ -27,10 +27,9 @@ registerRoutes([
 		pattern: '/',
 		view: (ctx) => html`<d2l-insights-adoption-dashboard-view
 							?demo=${ctx.options.isDemo}
```
Vitalii-Misechko, line 29: we don't need this line right?

Vitalii-Misechko (commented):
It seems that Course Access Count doesn't has loading spinner
image

demo/ad-demo-view.js
```diff
@@ -9,6 +9,12 @@ class AdDemoView extends LitElement {
 		// code related to visual diff tests
 		const hashMap = parseHash(window.location.hash);
 
+		const serviceWorker = hashMap.has('immediate') ? worker : waitingWorker;
+		serviceWorker.start();
+
+		// may be used by cards to hard-code things like axis ticks for consistency of visual diffs
+		document.documentElement.setAttribute('isDemo', true);
```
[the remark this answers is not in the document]
johngwilkinson: Maybe. This solution comes from engagement, where it was how we fixed a flaky visual diff test. I'm trusting that we ruled out other fixes then.

model/data.js
johngwilkinson: Right, the idea is to let mobx do it. Basically we have data sources which are being monitored by their rendering code: update the data source, and the ui redraws it.

johngwilkinson:
> It seems that Course Access Count doesn't has loading spinner image

Right, it just has stub data and doesn't load from the server yet (mocked or real) - so no spinner is expected yet.

Vitalii-Misechko (approved):
LGTM!

components/cards/system-access.js
```diff
@@ -237,19 +206,24 @@ class SystemAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+			marker: {
 				enabled: data && data.length === 1
 			},
 			color: 'var(--d2l-color-celestine)'
 		}];
 	}
+
+	get data() {
+		// possibly the server will return timestamps and not be nested, in which case this can be simplified
+		return this.resultSet.data.systemAccessData?.map(([d, v]) => [new Date(d).getTime(), v]) || [];
```
bemailloux, line 222: Is it ok for this value to be `undefined`? Or does it have to be an array?

bemailloux (approved):
LGTM!

PR #24, opened 2022-02-03
US132939: system access ui tweaks

johngwilkinson:
See thread: https://d2l.slack.com/archives/CMNGJ22G0/p1643900980413739

Quality Notes: just checked the demo for this
- PM approved changes
- tested w. Narrator and keyboard: seems to work quite well with scan either off or on

(see visual diff for image once that is available)


components/cards/system-access.js
```diff
@@ -218,8 +211,13 @@ class SystemAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 
 	get data() {
-		// possibly the server will return timestamps and not be nested, in which case this can be simplified
-		return this.resultSet.data.systemAccessData?.map(([d, v]) => [new Date(d).getTime(), v]) || [];
+		// possibly the server will return timestamps and not be nested, in which case this may need to adjust
+		// for the server timezone
+		return this.resultSet.data.systemAccessData?.map(([d, v]) => {
+			const [year, month, day] = d.split('-');
+			// midnight, browser time
+			return [new Date(year, month - 1, day, 0, 0, 0).getTime(), v];
```
NicholasHallman, line 219: year, month and day are string here and then we subtract one from the month. I know that js type coersions will result in a number for month but should we just explicitly map the split results to numbers?

NicholasHallman (approved):
Vis diff and slack images look great. LGTM

components/cards/system-access.js
johngwilkinson: 🤷 I don't know if that's better or not. This way is less code I suppose.

components/cards/system-access.js
```diff
@@ -63,10 +63,19 @@ class SystemAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 			<div class="d2l-insights-system-access-title d2l-skeletize d2l-skeletize-45 d2l-body-standard">${this._cardTitle}</div>
 			<d2l-labs-chart
 				.options="${this._chartOptions}"
+				.globalOptions="${this._globalOptions}"
 				?skeleton="${this.resultSet.isLoading}"
 			></d2l-labs-chart>`;
 	}
 
+	get _globalOptions() {
```
bemailloux, line 71: How global is "global"? Is it restricted to this card or does it go beyond that?

bemailloux (approved):

components/cards/system-access.js
johngwilkinson: I think it may affect others. This is probably okay (in fact, we may want to move this setting into chart.js at some point), but we'll have to keep it in mind.

PR #26, opened 2022-02-07
US132940 -  [Adoption] Add Course Access Stacked Bar Chart

Sean-Kurian:
Quality Notes

- Functional
    -  End to End testing after API changes

- UX 
  https://miro.com/app/board/o9J_lqOQbyc=/ -> Demo done

- Cost
    - Query for data

- Performance/Security/Dev Ops/Documentation
    - N/A


components/cards/course-access-counts-card.js
```diff
@@ -161,16 +168,20 @@ class CourseAccessCountsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-		const data = this._data;
-		return [{
-			name: this._yAxisTitle,
-			data,
-			marker:{
-				enabled: data && data.length === 1
-			},
-			color: 'var(--d2l-color-celestine)'
-		}];
+		const data = this.data;
```
johngwilkinson, line 171: might as well get rid of this method and just reference `this.data` directly instead

components/cards/course-access-counts-card.js
```diff
@@ -82,16 +79,29 @@ class CourseAccessCountsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-		</div>`;
+			<d2l-labs-chart
+				.options="${log(this._chartOptions)}"
+				.globalOptions="${this._globalOptions}"
+				?skeleton="${this.resultSet.isLoading}"
+			></d2l-labs-chart>`;
 	}
 
-	get chartOptions() {
+	get _globalOptions() {
```
johngwilkinson, line 96: Likely not needed for this chart because it doesn't display dates

components/cards/course-access-counts-card.js
```diff
@@ -1,51 +1,48 @@
...
+import { computed, decorate } from 'mobx';
 import { css, html } from 'lit-element/lit-element.js';
 import { BEFORE_CHART_FORMAT } from '../chart/chart';
 import { bodyStandardStyles } from '@brightspace-ui/core/components/typography/styles';
 import { Localizer } from '../../locales/localizer';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
+import { UserTrendColorsIterator } from '../consts.js';
+
+const log = (v) => {
```
johngwilkinson, line 11: If you're keeping this in, maybe just throw on a comment about what it's for - it took me a minute

test/server.js
```diff
@@ -68,6 +68,24 @@ const systemAccessDemoData = [
 	]
 ];
 
+const courseAccessDemoData = [{
+	name: 'Student',
+	data: [100, 30, 20, 40, 150],
```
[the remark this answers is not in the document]
johngwilkinson: This is should be an exact example of what the server should return, though. Right now, you've got "course 1", "course 2" etc. hard-coded, so those will appear regardless of the server data. Possibly the server will need to return an object with the series and the labels separately - I don't remember exactly what Highcharts will take.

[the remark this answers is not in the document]
NicholasHallman: I think we should keep conversations "unresolved" until the reviewer / implementer feels confident that the comments have been fufilled. Either because the implementer added the suggestion requested by the reviewer or the reviewer, through conversation with the implementer, decided that the change isn't necesarry.

NicholasHallman:
`FYI` You can safely ignore visual diff tests untill your feature is completed. Make sure you review the report generated by the bot PR so you can verify that it captured the feature properly

test/server.js
```diff
@@ -91,6 +91,25 @@ const systemAccessDemoData = [
...
+}, {
+	name: 'Administrator',
+	data: [50, 60, 70, 80, 90],
+	color: '#008EAB'
+}, {
+	name: 'Instructor',
+	data: [90, 100, 110, 130, 150],
+	color: '#E87511'
+}],
+labels: ['Science', 'Math', 'Art', 'Physics', 'Comp. Sci'] };
```
johngwilkinson, line 111: minor: to test long names, can you give one of these a longer name like "Faculty of unreasonably long names that exist only for testing purposes"

components/cards/course-access-counts-card.js
```diff
@@ -1,51 +1,29 @@
 import 'highcharts';
+import { computed, decorate } from 'mobx';
 import { css, html } from 'lit-element/lit-element.js';
 import { BEFORE_CHART_FORMAT } from '../chart/chart';
 import { bodyStandardStyles } from '@brightspace-ui/core/components/typography/styles';
 import { Localizer } from '../../locales/localizer';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin.js';
+//import { UserTrendColorsIterator } from '../consts.js';  -> used to dynamically generate correct sequence of colours with server data
```
johngwilkinson, line 9: remove if unused

components/cards/course-access-counts-card.js
```diff
@@ -1,51 +1,29 @@
...
+
+const log = (v) => {
```
johngwilkinson, line 11: remove now?

components/cards/course-access-counts-card.js
```diff
@@ -141,36 +122,29 @@ class CourseAccessCountsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-			color: 'var(--d2l-color-celestine)'
-		}];
+	get data() {
+		return this.dataSet.data;
 	}
 
 }
+
+decorate(CourseAccessCountsCard, {
+	data: computed
```
johngwilkinson, line 148: This isn't wrong, but since the getter doesn't do any substantial work, there's no advantage to this decoration

components/consts.js
```diff
@@ -0,0 +1,48 @@
+export const USER_TREND_COLORS = [
```
johngwilkinson, line 1: Is this used?

model/dataSets/courseAccessDataSet.js
```diff
@@ -0,0 +1,12 @@
+import { DataSet } from '../dataSet';
+
+export class CourseAccessDataSet extends DataSet {
+	process(data) {
+		if (typeof data.courseAccessData !== 'undefined') {
```
johngwilkinson, line 5: `undefined` and `null` are both falsey, so your ternary expressions that reference this don't care about the difference. So this whole function can simply be `return data.courseAccessData;`

components/cards/course-access-counts-card.js
```diff
@@ -141,36 +122,29 @@ class CourseAccessCountsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 				enabled: false,
 			},
 			legend: {
-				enabled: false,
+				enabled: true,
```
johngwilkinson, line 125: I believe this chart is not supposed to have a legend: it is sharing the roles legend that appears down below tool usage by role

Sean-Kurian: So will course access be on the same row as tool usage then?

johngwilkinson (approved):
My only concern is the colours - possibly on other charts too. I seem to recall there are a specific set of colours Kevin selected for us, but I'm not sure how that's supposed to work. Seems like something for a separate PR in any case.

bemailloux (approved):

PR #46, opened 2022-02-22
Us133252/csv export

johngwilkinson:
This adds csv export which creates a .zip from all dataSets which are marked for export. So far, only System Access is actually exported.

Note: switched to csv-stringify from export-from-json because it saves having to construct full json objects with header names that just get converted back to plain csv rows again. We also no longer can use the save-file functionality of export-from-json because we need to build a .zip from all the csvs.

Quality Notes
- functional
  - demo page: export and check csv; also check with multiple files in csv
  - [x] e2e
- performance: n/a
- security: n/a
- devops: n/a
- ux/docs: same button as other dashboards


components/cards/system-access.js
```diff
@@ -211,13 +211,7 @@ class SystemAccessCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 
 	get data() {
-		// possibly the server will return timestamps and not be nested, in which case this may need to adjust
-		// for the server timezone
-		return this.resultSet.data.systemAccessData?.map(([d, v]) => {
-			const [year, month, day] = d.split('-');
-			// midnight, browser time
-			return [new Date(year, month - 1, day, 0, 0, 0).getTime(), v];
-		}) || [];
+		return this.dataSet.data;
```
Vitalii-Misechko, line 214: nice!

model/dataSet.js
```diff
@@ -18,6 +18,29 @@ export class DataSet {
 		return data;
 	}
 
+	get isExport() {
+		return false;
+	}
```
Vitalii-Misechko, line 23: Should we move all these new members to a mixin?

model/dataSets/systemAccessDataSet.js
```diff
@@ -0,0 +1,16 @@
...
+			// midnight, browser time
+			return [new Date(year, month - 1, day, 0, 0, 0).getTime(), v];
+		}) || [];
+	}
+
+	get isExport() { return true; }
+	processForExport(data) { return data.systemAccessData || []; }
+	get headersForExport() { return ['export:date', 'export:systemAccessCount']; }
+	get exportFilenameTerm() { return 'export:systemAccessFilename'; }
+}
```
Vitalii-Misechko, line 16: neat!

model/data.js
```diff
@@ -12,11 +13,15 @@ export class Data {
 		this._resultSets.forEach(resultSet => resultSet.load());
 	}
 
-	getResultSet(name) {
-		return this._resultSets.find(r => r.name === name);
+	getResultSet(resultSet) {
```
Vitalii-Misechko, line 16: `resultSet` => `resultSetInfo`?

model/resultSets.js
```diff
@@ -1,6 +1,8 @@
 export const resultSets = {
```
Vitalii-Misechko, line 1: `resultSets` => `resultSetInfos`?

model/data.js
```diff
@@ -12,11 +13,15 @@ export class Data {
...
+		return this._resultSets.find(r => r.name === resultSet.name);
 	}
 
 	getDataSet(name) {
 		return this._dataSets.get(name);
 	}
+
+	async export(localize) {
+		await exportToZip(localize('export:filename'), [...this._dataSets.values()].filter(d => d.isExport), localize);
+	}
```
Vitalii-Misechko, line 26: `thought` It feels like this implementation can be a part of `adoption-dashboard-view.js` or a separate class that accesses `_dataSets` via getter. But it's fine here I guess

Vitalii-Misechko (approved):
LGTM!

locales/en.js
```diff
@@ -22,5 +23,11 @@ export default {
...
 
 	"toolUsageTrendCard:title": "Tool Usage Trend by Tool",
-	"toolUsageTrendCard:yAxisTitle": "Usage Count"
+	"toolUsageTrendCard:yAxisTitle": "Usage Count",
+
+	"export:filename": "Adoption Dashboard",
+	"export:date": "Date",
+	"export:systemAccessCount": "System Access Count",
+	"export:systemAccessFilename": "Adoption Dashboard - 'System Access'",
+	"export:toolUsageTrendFilename": "Adoption Dashboard - 'Tool Usage Trend by Tool'"
```
bemailloux, line 32: Should the file names be more "computer friendly" (like camel or snake case)?

bemailloux (approved):

locales/en.js
johngwilkinson: These names were specified by PM. Do you have specific concerns about them that I should take to her?

bemailloux: Just thinking about that one lab computer every institution has that's still running XP or older. But usually those are "off-the-grid" these days :)

PR #52, opened 2022-02-24
US133198: [Adoption] Add Enrollments and Withdrawals Trends Double Line Chart 

bbabu01:
[US133198](https://rally1.rallydev.com/#/?detail=/userstory/613171541653&fdp=true): [Adoption] Add Enrollments and Withdrawals Trends Double Line Chart to Enrollments and Withdrawals Section

- Functional Testing
   - [x] Demo page using stub data
- Security, devops, perf: N/A
- UX and docs
   -  [ ] Demo
image


components/chart-legend.js
```diff
@@ -1,14 +1,15 @@
...
 import { Localizer } from '../locales/localizer';
 import { MobxLitElement } from '@adobe/lit-mobx';
 import { SkeletonMixin } from '@brightspace-ui/core/components/skeleton/skeleton-mixin';
 
 class ChartLegend extends SkeletonMixin(Localizer(MobxLitElement)) {
 	static get properties() {
 		return {
-			dataSet: { type: Object, attribute: false }
+			dataSet: { type: Object, attribute: false },
+			color: { type: Array, attribute: false }
```
Vitalii-Misechko, line 12: I would suggest a `colorsOverrid` (colors-override) property. If filled then it is used instead of component's built-in colors
```
const colors = this.colorsOverride ? this.colorsOverride : UserTrendColorsIterator(0, 1, this._data.length);
```

components/cards/enrollments-withdrawals-trends.js
```diff
@@ -0,0 +1,228 @@
...
+			},
+			color: colors[name]
+		}), this) || [];
+	}
+
+	getData(usageData) {
+		return usageData?.map(([d, v]) => {
+			const [year, month, day] = d.split('-');
+			// midnight, browser time
+			return [new Date(year, month - 1, day, 0, 0, 0).getTime(), v];
```
Vitalii-Misechko, line 221: I believe that you can use `formatDateDataPoints` function

Vitalii-Misechko (commented):
I've added a few minor suggestions

components/chart-legend.js
bbabu01: Thanks for the suggestion. Since I will be overriding with an array of 2 colors for this card, shouldn't the components built in colors also be an array instead of a generator?

`const colors = this.colorsOverride ? this.colorsOverride : [...UserTrendColorsIterator(0, 1, this._data.length)];`

Vitalii-Misechko: `[...UserTrendColorsIterator(0, 1, this._data.length)]` this is array. And we need to use the generator to cover the case when the number of legend items is more than colors available. The generator starts from the first color when all colors are taken from `USER_TREND_COLORS`

bemailloux (commented):
Looks good! Just waiting for implementation of Vitalii's suggestions

Vitalii-Misechko (dismissed):
LGTM!

If this (the legend too far from the chart) okay to Megan then I'm fine too
image

Vitalii-Misechko (approved):

model/dataSets/enrollmentsWithdrawalsTrendDataSet.js
```diff
@@ -0,0 +1,17 @@
+import { DataSet } from '../dataSet';
+
+export class EnrollmentsWithdrawalsTrendDataSet extends DataSet {
+
+	process(data) {
+		return [{ name:'Enrollments', trendData: data?.enrollmentsWithdrawalsData?.Enrollments || [] }, { name:'Withdrawals', trendData: data?.enrollmentsWithdrawalsData?.Withdrawals || [] }];
```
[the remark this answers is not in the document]
bbabu01: Not this name. But to your point, it displays the ones from EnrollmentsWithdrawalsLegendDataSet in the legend, which is currently not localized. Don't see a straightforward way to do this since we need it to be in a dataset to pass to chart-legend.js. For now I will hardcode the localized labels in  adoption-dashboard-view.js

components/cards/enrollments-withdrawals-trends.js
```diff
@@ -0,0 +1,216 @@
...
+					}
+				}
+			},
+
+			series: this._series
+		};
+	}
+
+	get _series() {
+		if (!Object.values(this.dataSet.data).length) return [];
```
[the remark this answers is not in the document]
bbabu01: thanks, fixed it. Was using a map earlier.

model/dataSets/enrollmentsWithdrawalsTrendDataSet.js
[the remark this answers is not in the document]
Vitalii-Misechko: Thank you for that details. In your case, `EnrollmentsWithdrawalsLegendDataSet` can return a static array with langterms. You also need to improve `d2l-insights-chart-legend` in order it can apply `localize` if needed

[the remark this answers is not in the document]
Vitalii-Misechko: After a closer look, I see that `name` is used for a screen reader. Therefore, must be localized

image

You can introduce an object with lang term. Have a look at `langTerms` as example
```JS
		const colors = { Enrollments: ENROLLMENTS_TREND_COLOR, Withdrawals: WITHDRAWALS_TREND_COLOR };
		const langTerms = { Enrollments: 'term1', Withdrawals: 'term2' };

		return data.map(({ name, trendData }) => ({
			name: this.localize(langTerms[name]),
			data: formatDateDataPoints(trendData),
			marker: {
				enabled: trendData && trendData.length === 1
			},
			color: colors[name]
		}), this) || [];
```

components/cards/enrollments-withdrawals-trends.js
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	get _series() {
+		if (!this.dataSet.data.length) return [];
+
+		const data = this.dataSet.data;
+		const colors = { Enrollments: ENROLLMENTS_TREND_COLOR, Withdrawals: WITHDRAWALS_TREND_COLOR };
+
+		return data.map(({ name, trendData }) => ({
+			name: name,
```
Vitalii-Misechko, line 198: Should be localized

model/dataSets/enrollmentsWithdrawalsLegendDataSet.js
```diff
@@ -0,0 +1,11 @@
+import { DataSet } from '../dataSet';
+
+/**
+ * Returns [{id, name}]
+ */
+export class EnrollmentsWithdrawalsLegendDataSet extends DataSet {
+	process(data) {
+		//Sort keys to ensure labels order is always Enrollments, Withdrawals
+		return Object.keys(data?.enrollmentsWithdrawalsData ?? []).sort().map((name, index) => ({ id: index, name: name }));
```
Vitalii-Misechko, line 9: It could return a static array of langterms. These langterms could be localized in `d2l-insights-chart-legend`. For instance, you could introduce a new boolean property `langterms`. When this property is specified for `d2l-insights-chart-legend` then for the component calls `this.localize` for each legend item

Vitalii-Misechko (commented):
It seems the legend is too close to the chart. Should you add any additional padding below the chart?

image

components/chart-legend.js
```diff
@@ -98,7 +100,7 @@ class ChartLegend extends SkeletonMixin(Localizer(MobxLitElement)) {
 				class="d2l-skeletize d2l-insights-chart-legend-item d2l-body-compact"
 			>
 				<div class="d2l-insights-chart-legend-color" style="background-color:${color};"></div>
-				<p class="d2l-insights-chart-legend-name">${legendItem.name}</p>
+				<p class="d2l-insights-chart-legend-name">${this.shouldLocalize && !this.skeleton ? this.localize(legendItem.name) : legendItem.name}</p>
```
bbabu01, line 103: @Vitalii-Misechko had to add check for !this.skeleton here, otherwise legends color image was displayed during loading

model/dataSets/enrollmentsWithdrawalsTrendDataSet.js
```diff
@@ -0,0 +1,20 @@
...
+			{ name:'Enrollments', trendData: data?.enrollmentsWithdrawalsData?.Enrollments || [] },
+			{ name:'Withdrawals', trendData: data?.enrollmentsWithdrawalsData?.Withdrawals || [] }
+		];
+	}
+
+	get isExport() { return true; }
+	processForExport(data) {
+		const enrollments = data?.enrollmentsWithdrawalsData?.Enrollments || [];
+		const withdrawals = data?.enrollmentsWithdrawalsData?.Withdrawals || [];
+		return enrollments.map(e => [e[0], e[1], withdrawals.find(w => e[0] === w[0])[1]]);
```
bemailloux, line 16: Wow this line is... complicated. Could you explain what's happening here? Or leave a comment in code?

model/dataSets/enrollmentsWithdrawalsLegendDataSet.js
```diff
@@ -0,0 +1,19 @@
...
+		return data ? [
+			{
+				'id': 0,
+				'name': 'export:enrollments'
+			},
+			{
+				'id': 1,
+				'name': 'export:withdrawals'
+			}
+		] : {};
```
bemailloux, line 17: This returns different datatypes if data exists or not - it should probably return the same datatype

components/chart-legend.js
```diff
@@ -11,7 +11,9 @@ class ChartLegend extends SkeletonMixin(Localizer(MobxLitElement)) {
 		return {
 			dataSet: { type: Object, attribute: false },
 			title: { type: String, attribute: true },
-			label: { type: String, attribute: true }
+			label: { type: String, attribute: true },
+			colorsOverride : { type: Array, attribute: false },
+			shouldLocalize : { type: Boolean, attribute: true }
```
bemailloux, line 16: Just wondering, why is this an option? Shouldn't this always be `true`?

bemailloux (commented):
Looking good, just a few small things

model/dataSets/enrollmentsWithdrawalsTrendDataSet.js
bbabu01: This was to merge enrollments and withdrawals arrays to a single array by comparing first columns(date) of each. Initially I did something like 
`enrollments.map((e, index) => [ e[0], e[1], withdrawals(index)[1] ] );
`
Since this involves an assumption that both enrollments and withdrawals arrays will have dates  in same order and value, went with comparison instead. I think that assumption might be good for this scenario though.

Enrollments `[[date1, c1],[date2, c2].....] `
Withdrawals `[[date1, d1],[date2, d2].....] `
to 
`[[date1, c1, d1], [date2, c2, d2].....] `

components/chart-legend.js
bbabu01: Legend texts for other cards are dynamic right. So we will have to explicitly add translations for all possible legend values for other cards if this is always true.

model/dataSets/enrollmentsWithdrawalsTrendDataSet.js
Vitalii-Misechko: Thank you for your explanation. Could you compose comments from the last four lines of your post? I believe I will not recall what the code does in a week :)

Vitalii-Misechko (approved):
LGTM! @bbabu01, thank you for your improvements!

Minor: 
- [ ] remove useless test

Vitalii-Misechko (approved):

PR #55, opened 2022-02-25
US135365 [Adoption] Add Tool Usage Trend by Role Stacked Area Chart to Tools Usage

Vitalii-Misechko:
[US135365](https://rally1.rallydev.com/#/?detail=/userstory/622671113099&fdp=true): [Adoption] Add Tool Usage Trend by Role Stacked Area Chart to Tools Usage

* Functional Testing 
  * [x] Lit 2 Phase 1 compliant changes
  * [x] Accessibility
  * [x] RTL
  * [x] Responsive
  * [x] E2E Test
* Security, devops, perf: N/A
* UX and docs

* [ ] Demo


pages/adoption-dashboard-view.js
```diff
@@ -160,35 +170,56 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+
+						<d2l-insights-chart-legend
+							title="${this.localize('toolsUsageSection:tools')}"
+							label="${this.localize('toolsUsageSectionA11y:tools')}"
+							.dataSet="${toolsLegendDataSet}"
+						></d2l-insights-chart-legend>
+					</div>
 
-					<div class="d2l-insights-card">
+					<div class="d2l-insights-card-with-legend">
```
johngwilkinson, line 199: If the legend is appearing with the card, why not use Highcharts built-in legend?

Vitalii-Misechko: I copied the solution of the Engagement Dashboard. For some reason, we decided to use our custom legend

Vitalii-Misechko: Should I try to implement the legend with Highcharts built-in capabilities?

johngwilkinson: We did that in Engagement because we had one legend shared by two or three charts. If the design is to have a legend for each chart, then we should most likely use Highcharts legend, since it should be very little code (hopefully!)

Vitalii-Misechko: I'll check it out

Vitalii-Misechko: @johngwilkinson 

Essentially, the built-in chart legend is a lot easier to implement. But it is a part of Highcharts chart, therefore it affects the size of card. Because the number of items in a legend is unbound I think the custom build legend would allow a better design solution. What do you think?

image

johngwilkinson: Interesting! I wonder if we could allow those cards to grow vertically together depending on the size of the legend. Probably complicated, but perhaps a little css would do it. If not, then maybe having the legend outside the card is best. I'd have to defer to Megan on this of course.

johngwilkinson (approved):

NicholasHallman (approved):

components/cards/tool-usage-trend-card.js
```diff
@@ -198,20 +200,24 @@ class ToolUsageTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-		if (!this.dataSet.data.length) return [];
+		if (this._isEmpty) return [[new Date(), 0]]; // to show axises
 
 		const data = this.dataSet.data;
 		const colors = [...UserTrendColorsIterator(0, 1, data.length)];
 
-		return data.map(({ toolId, toolName, usageData }) => ({
-			id: toolId,
-			name: toolName,
+		return data.map(({ id, name, usageData }) => ({
```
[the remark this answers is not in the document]
Vitalii-Misechko: Good point! Actually, I have tried to do so. But representing date as an integer fails CI because of different timezone. I prefer this option because highcharts specific logic is easier to understand here

PR #58, opened 2022-03-02
Us132942/tool usage comparison

bemailloux:
Adds tool usage comparison chart

image

On Draco LMS
scroll

Quality notes
* Functionality (tested local & draco stack)
  * Shows correct aggregations per tool / per org unit (includes data from the past 15 days, today inclusive)
  * Shrinks / expands horizontal axis based on data count
  * Org unit names are fully shown (including on multiple lines) if there is enough room vertically to show all org units. Otherwise, the chart becomes scrollable and org unit names are truncated to a single line
  * Tooltips show org unit name, id, tool name and count
  * Tool colours match the legend and tool usage chart
  * Shows org units which have no data. In a future release we may modify this functionality to "group" no data org units into an expandable entry.
  * NO export: I've made a separate task for this
* Perf / large data cases
  * Aggregation should be O(n) (but loops over rows multiple times)
  * Handles large amounts of org units & long org unit names
* UX
  * Responsive
  * Will focus on accessibility in a separate task
* Docs, devops, security: N/A


Vitalii-Misechko:
Have you had a chance to see this design? I believe it is latest

image

bemailloux:
> Have you had a chance to see this design? I believe it is latest

I have not seen this design. Is there a link to it somewhere?

Vitalii-Misechko:
> > Have you had a chance to see this design? I believe it is latest
> 
> I have not seen this design. Is there a link to it somewhere?

I've received it from Megan. Sorry, I don't have the link

components/cards/shared-styles.js
```diff
@@ -0,0 +1,24 @@
+import { css } from 'lit-element/lit-element.js';
+
+export const cardSharedStyles = css`
```
Vitalii-Misechko, line 3: nice!

model/dataSets/toolUsageComparisonDataSet.js
```diff
@@ -1,8 +1,95 @@
...
+		const toolInfoSorted = sortBy(
+			Object.entries(toolInfo)
+				.map(([id, name]) => ({ id, name }))
+				.filter(tool => dataToolsSet.has(tool.id))
+			, 'name'
+		);
+
+		// step 1: filter out rows outside of the period of interest
+		const startDay = subDays(dateRange.today, LOOK_BACK_PERIOD_DAYS);
+		rows = rows.filter(r => r.D >= startDay);
```
Vitalii-Misechko, line 53: Do we compare strings? If yes should we compare dates?

test/model/dataSets/toolUsageComparisonDataSet.test.js
```diff
@@ -0,0 +1,99 @@
...
+				toolComparisonData : []
+			});
+		});
+
+		it('transforms data', () => {
+			// filters out data that's older than 2 weeks old
+			// includes a tool entry for each tool that was present in the original dataset (even if that data was filtered out)
+			// aggregates correctly
+			// aggregate counts are in order, based on org unit name
+			// filters out root OU
```
Vitalii-Misechko, line 23: Thank you for these comments!

Vitalii-Misechko (dismissed):
LGTM!

model/dataSets/toolUsageComparisonDataSet.js
bemailloux: We are comparing strings here. This will work correctly as long as date strings are in 'YYYY-MM-DD' format (we won't get anything weird like "2 > 10" because we always use "02" instead). At the resultset level I don't expect the format will ever change, so I'm ok with leaving it as is.

Vitalii-Misechko (approved):

Vitalii-Misechko:
`curious` Can we truncate the long course names to some reasonable number of characters?

image

Vitalii-Misechko:
`nit` broken styles

image

Vitalii-Misechko:
`responsivness`  Should the card shrink to the regular card width at some breakpoint?

image

bemailloux:
> Can we truncate the long course names to some reasonable number of characters?
> Should the card shrink to the regular card width at some breakpoint?

Probably it's better if I ask these questions in a demo and make changes in another PR if needed

model/dataSets/toolUsageComparisonDataSet.js
```diff
@@ -1,8 +1,95 @@
...
+		// TODO: centralize legend / colours (remove duplication)
+		const dataToolsSet = new Set(rows.map(r => r.t));
+		const toolInfoSorted = sortBy(
+			Object.entries(toolInfo)
+				.map(([id, name]) => ({ id, name }))
+				.filter(tool => dataToolsSet.has(tool.id))
+			, 'name'
+		);
+
+		// step 1: filter out rows outside of the period of interest
```
johngwilkinson, line 51: this should already have happened on the server

johngwilkinson (approved):

model/dataSets/toolUsageComparisonDataSet.js
bemailloux: Is this kind of a general statement? For any filter change, should we expect to be going to the server and filtering there? (plus or minus some optimizations maybe?)

johngwilkinson: I think we can do role and tool filtering client-side, at least mostly. Org unit and date filter changes I'm assuming go to the server, because we generally won't have the data in the client. For org units, I think that's important because there could be a huge number of them given the persona for this dashboard. For dates, though, I suppose we could pre-load a larger range to try and avoid server calls. That seems like an optimization we likely just don't need though.

bemailloux: Makes sense to me. Thanks for clarifying :)

(Will fix up the front-end after working on back-end changes)

PR #65, opened 2022-03-03
Error Handling

NicholasHallman:
Functional
 - Unit Tests
   - [X] isQueryError dataset will show an alert on the card and hide the chart
   - [X] All result sets failing will show an error on the dashboard and hide the view
   - [X] alert shows for system access trend when access data set fails 
 - VisDiff
   - [x] Chart error message
   - [x] All error message

 - E2E
   - [x] Error shows when theres a 500 from server

UI / UX
 - [x] demo

Accessibility
 - Toast Evens to indicate some data didn't arrive
 - Toast Event when not enough data arrived to draw the dashboard
 - Tabbing into a card with an error alert reads the card name and the error message
 
Docs / DevOps / Security - NA

1HsRwvI9Bc


demo/index.html
```diff
@@ -5,6 +5,7 @@
 		<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
 		<script type="module" src="/node_modules/@brightspace-ui/core/components/demo/demo-page.js"></script>
 		<script type="module" src="/demo/ad-demo-view.js"></script>
+		<script type="module" src="/demo/commands.js"></script>
```
NicholasHallman, line 8: Include the testing commands in demo so the file isn't rolled up in prd

johngwilkinson:
I see conflicts ⚠️ 

For quality notes: "functional" means "functional testing" - so, how you tested and what cases you tested. (For this, because it's in response to server errors, I would be looking for e2e testing.)

model/data.js
```diff
@@ -6,11 +7,18 @@ export class Data {
...
 	}
 
 	load({ orgUnitIds, userSettings }) {
+		this._allError = false;
+
 		this._dataApiClient.startDataRequest({ orgUnitIds, userSettings });
-		this._resultSets.forEach(resultSet => resultSet.load());
+		const resultPromises = this._resultSets.map(resultSet => resultSet.load());
+
+		Promise.all(resultPromises).then(results => {
```
johngwilkinson, line 19: nit: preferred idiom now would be
```
const results = await Promise.all(...);
if (results...) ...;
```

johngwilkinson (commented):
Looks like a visual diff or two would be nice (maybe another PR if that's tricky)

components/cards/tool-usage-trend-card.js
```diff
@@ -68,6 +71,13 @@ class ToolUsageTrendCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 			></d2l-labs-chart>`;
 	}
 
+	renderError() {
+		return html`
+			<div class="d2l-insights-system-access-title d2l-skeletize d2l-skeletize-45 d2l-body-standard">${this._cardTitle}</div>
+			<d2l-alert status="error">${this.localize('card:Error')}</d2l-alert>
+		`;
+	}
```
Vitalii-Misechko, line 79: `curious` Have you considered implementing error logic at `d2l-labs-chart` level?

model/data.js
```diff
@@ -25,3 +33,7 @@ export class Data {
 		await exportToZip(localize('export:filename'), [...this._dataSets.values()].filter(d => d.isExport), localize);
 	}
 }
+
+decorate(Data, {
+	_allError: observable
+});
```
Vitalii-Misechko, line 39: `alternative` I don't know if it works though

```JS
get allError() {
  return this._resultSets.every(r => !r.isQueryError); // result IsError should become observable
}

decorate(Data, {
	allError: computed;
});
```

demo/commands.js
```diff
@@ -0,0 +1,21 @@
...
+	return document.querySelector('d2l-insights-ad-demo-view')
+		.shadowRoot.querySelector('d2l-insights-adoption-dashboard');
+};
+
+const refresh = (dashboard) => dashboard.data.load(dashboard.userSettings);
+
+window.tests = {};
+
+window.errorStateTest = {
+	causeSystemAccessDataError: () => {
```
Vitalii-Misechko, line 12: Who calls this function? Could you point out the file?

NicholasHallman: These functions are for the demo, you can call them in the console to trigger an error state in the dashboard

NicholasHallman:
> I would be looking for e2e testing.

My plan for this is to have draco2 bdp send back 500 errors, one for each data set

model/data.js
NicholasHallman: It will work because isQueryError in the result sets are observables but I'm leaning towards keeping it like this because IMO if a variable is labeled `observable` then I know that if it changes it will cause a state change in the UI. Computed for me is more ambiguouse because it could mean that it's a large computation, not necesarely that it's suppose to update the state. Even though it could

components/cards/tool-usage-trend-card.js
NicholasHallman: Sure sounds good!

Vitalii-Misechko (dismissed):
LGTM!

model/data.js
johngwilkinson: Maybe this will clarify: MobX only cares about `observable` values and reacts to their changes: if an action derefrences an observable, that action will be run when the observable changes. Something being `computed` doesn't change anything about this: it simply tells mobx to cache the results of the function. So `computed` only does something if the function a) is called in an action (like `render()`), directly or indirectly; and b) references an `observable`, directly or indirectly.

insights-adoption-dashboard.js
```diff
@@ -52,6 +52,7 @@ class AdoptionDashboard extends Localizer(MobxLitElement) {
 				--d2l-insights-adoption-card-margin-top: 10px;
 				--d2l-insights-adoption-card-margin-right: 12px;
 				display: block;
+				min-height: 100vh;
```
johngwilkinson, line 55: Is this the reason a number of visual diffs are vertically a bit larger now? Is the purpose of this change to make the diffs more consistent, or something else?

johngwilkinson (commented):
Other than needing a merge/resolve and a couple of stray lines from an earlier one, this LGTM

insights-adoption-dashboard.js
NicholasHallman: When the dashboard only renders the error alert the page scrunches up and you can see whatever color the browser puts behind the dashboard. This just ensures that the dashboard is always at least the height of your screen with no scroll bar.

johngwilkinson (approved):

johngwilkinson (approved):

demo/commands.js
```diff
@@ -0,0 +1,49 @@
...
+const getDashboardView = () =>
+	document.querySelector('body > d2l-insights-ad-demo-view')
+		.shadowRoot.querySelector('d2l-insights-adoption-dashboard')
+		.shadowRoot.querySelector('d2l-labs-toaster > d2l-insights-adoption-dashboard-view');
+
+const refresh = (dashboard) => dashboard.data.load({ ...dashboard.userSettings, alert: getDashboardView() });
+
+window.tests = {};
+
+const log = (v) => console.log(v) || v;
```
Vitalii-Misechko, line 15: Could you add `allow` linting-rule for this line/file?
```JS
/* eslint no-console: "allow" */
```

demo/commands.js
```diff
@@ -0,0 +1,49 @@
+
+const getDashboard = () =>
```
Vitalii-Misechko, line 2: Could you add some comments about usage (or even usage examples)?

model/data.js
```diff
@@ -12,11 +12,15 @@ function splitAtFirst(s, ch) {
 	return [s.substring(0, i), s.substring(i + 1)];
 }
 
+const any = (array, predicate) => {
```
Vitalii-Misechko, line 15: `curious` Why not use `some`? I assume that you just don't like the name choice of the JS team for the built-in predicate (namely  `some`)  btw, me too 😆

insights-adoption-dashboard.js
```diff
@@ -129,7 +130,7 @@ class AdoptionDashboard extends Localizer(MobxLitElement) {
 				dataSetFactories,
 				dataApiClient: new DataApiClient(this.metronEndpoint)
 			});
-			this._data.load({ userSettings: this.userSettings });
+			this._data.load({ userSettings: this.userSettings, alert: this });
```
Vitalii-Misechko, line 133: Could you help me to figure out which method is called here? I see that Data class in  `data.js` has `load` method only with one parameter `load({ alert })`. Why do preserve `userSettings` then?

model/resultSets.js
```diff
@@ -6,7 +6,7 @@ export const resultSets = {
 		name: 'systemAccessData'
 	},
 	COURSE_ACCESS: {
-		name: 'courseAccessData'
+		name: 'courseAccessData',
```
Vitalii-Misechko, line 9: we don't need this comma, right?

test/visual-diff/adoption-dashboard.visual-diff.js
```diff
@@ -49,5 +49,38 @@ const VisualDiff = require('@brightspace-ui/visual-diff');
...
+				});
+			});
+
+			[
+				'causeSystemAccessDataError',
+				'causeToolUsageTrendDataError',
+				'causeCourseAccessDataError',
+				'causeEnrollmentDataError',
+				'causeAll'
+			].forEach(errorCommand => {
```
Vitalii-Misechko, line 73: Neat!!!! 🚀

components/cards/course-access-counts-card.js
```diff
@@ -23,6 +23,7 @@ class CourseAccessCountsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 	static get properties() {
 		return {
 			dataSet: { type: Object, attribute: false },
+			title: { type: String, attribute: true },
```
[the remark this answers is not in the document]
NicholasHallman: When the title is in the element the title gets read out by the screen reader when the error is highlighted. Also helps with speech navigation

components/cards/enrollments-withdrawals-trends.js
```diff
@@ -15,7 +15,8 @@ class EnrollmentsWithdrawalsTrendCard extends SkeletonMixin(Localizer(MobxLitEle
 
 	static get properties() {
 		return {
-			dataSet: { type: Object, attribute: false }
+			dataSet: { type: Object, attribute: false },
+			title: { type: String, attribute: true, reflect: true }
```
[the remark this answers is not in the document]
NicholasHallman: I use reflection to populate that property at render which helps with accessibility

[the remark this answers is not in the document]
Vitalii-Misechko: Can we replace it with `d2l-offscreen`? If not could you add a comment near `if (this.title === undefined)` to explain the trick. It just looks hacky.

model/data.js
NicholasHallman: I just didn't realize that's what `some` did LOL

insights-adoption-dashboard.js
NicholasHallman: That's an old call signature I'll fix that

model/data.js
```diff
@@ -25,9 +29,20 @@ export class Data {
...
+
 		this._dataApiClient.startDataRequest({ parentId: this.parentId });
-		this._resultSets.forEach(resultSet => resultSet.load());
+		const resultPromises = this._resultSets.map(resultSet => resultSet.load());
+
+		Promise.all(resultPromises).then(results => {
+			if (results.every(r => !r.result)) {
+				this._allError = true;
+			} else if (any(results, r => !r.result) && alert !== undefined) {
+				ToastEvent.dispatch(alert, alert.localize('dashboard:SomeErrorAxe'), alert.localize('dashboard:SomeErrorAxe'), false);
```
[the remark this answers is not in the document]
NicholasHallman: It was unfortunatly easier to place here because the data layer recognizes the error only once, when the error is processed. The presentation layer processes the data everytime there's an update which means that the alerts fire multiple times. I guess we could track the change in error states in the dashboard as a solution?

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Don't bother finding conflicts for ci vis diff
+test/visual-diff/screenshots/ci/golden/**/* merge=ours
```
NicholasHallman, line 2: Hopefully this makes future merging easier

model/data.js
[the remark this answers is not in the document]
Vitalii-Misechko: I see. I guess the proposed solution could work. Another alternative is Mobx-fying error state and using it in `reaction`

Vitalii-Misechko (approved):
LGTM!

PR #70, opened 2022-03-07
Us133252/tool usage export

johngwilkinson:
Quality Notes
- functional
  - demo page
  - e2e: chrome, ff, edge
- security/perf/devops: n/a
- ux/docs
  - story doesn't ask for ou/role names, but we could add them


model/dataSets/toolUsageTrendByRoleDataSet.js
```diff
@@ -37,4 +37,13 @@ export class ToolUsageTrendByRoleDataSet extends DataSet {
 			};
 		});
 	}
+
+	// This DataSet exports all the tool usage data (any one of the tool usage DataSets could do this)
```
bemailloux, line 41: Is it a problem if all the datasets do this?

bemailloux (approved):
LGTM!

model/dataSets/toolUsageTrendByRoleDataSet.js
johngwilkinson: Not sure if I understand the question, but I think yes? There will be one file in the export .zip for each `DataSet` with `isExport == true`, so we don't want the other two tool usage DataSets to do that.

Vitalii-Misechko (approved):

PR #71, opened 2022-03-07
US133252: course access export

johngwilkinson:
Adds csv export for Course Access, and also switches to use the `orgUnits` field of the response instead of `labels` - this is so we can later switch to labels like "Org Name (id: org id)".

Quality Notes
- functional: e2e along with incoming server-side changes: https://github.com/Brightspace/bdp-query/pull/190
- perf/security/devops: n/a
- ux/docs
  - need to check on desired labels
  - should also consider including role/ou names in export (story only asks for ids)


bemailloux (approved):
LGTM!

Vitalii-Misechko (approved):

PR #72, opened 2022-03-07
Us132942/tuc cleanup

bemailloux:
(1) Filter out non-root items from Tool Usage Trend / Tool Usage by Role
(2) Truncate org unit names in Tool Usage Comparison

Quality notes
* Testing
  * Tested on Draco LMS: 
    * TUT/TUR show correct aggregate of data from root org unit only
    * TUC shows correct aggregate of data from children only
* Security, UX, perf, devops, docs: N/A

image


model/dataSets/toolUsageTrendByRoleDataSet.js
```diff
@@ -9,9 +9,13 @@ export class ToolUsageTrendByRoleDataSet extends DataSet {
 		 * { "t": "11000", "o": "121535", "r": "579", "D": "2022-02-02", "count": 1 },
 		 */
 	process(data) {
+		// rows now includes data for root org unit and all its children. For this card we only want to show root data
 		const rows = data?.toolUsageTrendData ?? [];
+		const rootOrgUnitId = data?.lmsData?.rootOrgUnitId; // if this isn't provided, no data will appear
 
-		const toolsUsageDataMap = chain(rows)
+		const rowsForRootOrgUnit = rows.filter(r => r.o === rootOrgUnitId);
```
Vitalii-Misechko, line 16: I assume there is no filter by `rootOrgUnitId` at the backend side anymore, right?

test/server.js
```diff
@@ -173,13 +173,60 @@ export const toolUsageTrendData = {
...
+		{ t: '11000', r: '580', o: '6606', D: '2021-11-21', count: 30 },  { t: '11000', r: '580', o: '6606', D: '2021-11-22', count: 25 },
+		{ t: '11000', r: '580', o: '6606', D: '2021-11-21', count: 30 },  { t: '11000', r: '580', o: '6606', D: '2021-11-22', count: 25 },
+		{ t: '11000', r: '580', o: '6606', D: '2021-11-21', count: 30 },  { t: '11000', r: '580', o: '6606', D: '2021-11-22', count: 25 },
+		{ t: '11000', r: '580', o: '6606', D: '2021-11-23', count: 150 }, { t: '11000', r: '580', o: '6606', D: '2021-11-24', count: 151 },
+		{ t: '11000', r: '580', o: '6606', D: '2021-11-25', count: 130 }, { t: '11000', r: '580', o: '6606', D: '2021-11-26', count: 110 },
+		{ t: '11000', r: '580', o: '6606', D: '2021-11-27', count: 115 }, { t: '11000', r: '580', o: '6606', D: '2021-11-28', count: 151 },
+		{ t: '11000', r: '580', o: '6606', D: '2021-11-29', count: 130 }, { t: '11000', r: '580', o: '6606', D: '2021-11-30', count: 110 },
+		{ t: '11000', r: '580', o: '6606', D: '2021-12-01', count: 100 }, { t: '11000', r: '580', o: '6606', D: '2021-12-01', count: 15 },
```
Vitalii-Misechko, line 195: Should we add a block of rows for non-root org unit just to catch filter errors early?

Vitalii-Misechko (approved):
LGTM!

model/dataSets/toolUsageTrendByRoleDataSet.js
bemailloux: This is the filter that's used on the back-end:
```
arrayFilters: {
    o: [ lmsData.rootOrgUnitId, ...lmsData.orgUnits ]
},
```
In this case `lmsData.orgUnits` is the list of children of the root org unit id. Therefore entries for the root and all its direct children are included. 

In Tool Usage Trend and Tool Usage by Role, we only want to show the values for the root org unit (otherwise we would be double-counting), and in Tool Usage Comparison we only want to show the values for the children.

test/server.js
bemailloux: There are some: see line 184 (which is part of the same array)

components/cards/tool-usage-comparison-card.js
```diff
@@ -92,9 +92,14 @@ class ToolUsageComparisonCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-						fontFamily: 'Lato'
+						fontFamily: 'Lato',
+
+						// force axis labels to truncate if larger than width
+						textOverflow: 'ellipsis',
+						whiteSpace: 'nowrap',
+						width: '150px'
 					},
-					step: 1 // forces all x-axis elements to show (by default highcharts will hide some if there are too many)
+					step: 1, // forces all x-axis elements to show (by default highcharts will hide some if there are too many),
```
anhill-D2L, line 102: Nit: additional commas

model/dataSets/toolUsageTrendByRoleDataSet.js
Vitalii-Misechko: Thank you for that explanation.

PR #75, opened 2022-03-08
US133252: include role and ou names in export

johngwilkinson:
Quality: e2e on draco


Vitalii-Misechko (approved):

bemailloux (commented):
Is that "root OU doesn't have a name" issue getting fixed here or elsewhere?

johngwilkinson:
> Is that "root OU doesn't have a name" issue getting fixed here or elsewhere?

elsewhere (server-side)

bemailloux (approved):

PR #81, opened 2022-03-10
UI component of date range filter

Sean-Kurian:
[US132922](https://rally1.rallydev.com/#/?detail=/userstory/612182648205&fdp=true): [Adoption][Org Filter] Filter UI

* Functional Testing 
  * [x] Accessibility
  * [x] Responsive
  * [x] Test cases
    * [x] User will have the ability to open filter and select dates
    * [x] Date defaults to 14 day range, at most can be 6 months ago to present
* Security, devops, perf: N/A
* UX and docs

* [x] Demo (TBD)


pages/adoption-dashboard-view.js
```diff
@@ -159,7 +161,11 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 		const toolUsageComparisonDataSet = this.data.getDataSet(dataSets.TOOL_USAGE_COMPARISON);
 		const toolsLegendDataSet = this.data.getDataSet(dataSets.TOOL_USAGE_TREND_BY_TOOL_LEGEND);
 		const rolesLegendDataSet = this.data.getDataSet(dataSets.TOOL_USAGE_TREND_BY_ROLE_LEGEND);
-
+		const today = formatDate(new Date(), { format: 'yyyy-MM-dd' });
+		const twoWeeksAgo = formatDate(new Date(Date.now() - 12096e5), { format: 'yyyy-MM-dd' }); //magic number, 14 days in milliseconds
+		const sixMonthsAgo = new Date();
+		sixMonthsAgo.setMonth(sixMonthsAgo.getMonth() - 6);
```
bemailloux, line 167: Does this work when `now.getMonth() - 6` is negative?

pages/adoption-dashboard-view.js
```diff
@@ -159,7 +161,11 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 		const toolUsageComparisonDataSet = this.data.getDataSet(dataSets.TOOL_USAGE_COMPARISON);
 		const toolsLegendDataSet = this.data.getDataSet(dataSets.TOOL_USAGE_TREND_BY_TOOL_LEGEND);
 		const rolesLegendDataSet = this.data.getDataSet(dataSets.TOOL_USAGE_TREND_BY_ROLE_LEGEND);
-
+		const today = formatDate(new Date(), { format: 'yyyy-MM-dd' });
+		const twoWeeksAgo = formatDate(new Date(Date.now() - 12096e5), { format: 'yyyy-MM-dd' }); //magic number, 14 days in milliseconds
```
bemailloux, line 165: Alternatively, you could just write out 14\*24\*60\*60\*1000

bemailloux (commented):
~~You might want to ask for a design mockup from Kevin or Megan. I'd have expected a dropdown with the date ranges inside, in order to match the style of the org unit filter. Maybe they have other ideas.~~ Nvm looked at the wrong story

pages/adoption-dashboard-view.js
Sean-Kurian: It does but can potentially break on edge cases like month endings, where if today was March 31st and I tried to go back 1 month, it would return February 31st. I'm going to write a custom function to handle that.

Sean-Kurian: I misread the function initially, it does work. If it fell on February 31st, it would just return March 3rd instead (or March 2nd) depending on the year. It's just a different way to handle dates - I had initially written a custom function for it, but decided that was clunkier when theres no real definition of 6 months ago (180 days vs 182.5 days vs 6 calendar months on the same day).

Sean-Kurian:
Ignore the repeated merges of vis-diffs for now, trying to solve an issue where certain components like 'Learn More' and 'Export to CSV' arent being captured in the vis-diff 100% of the time.

pages/adoption-dashboard-view.js
```diff
@@ -192,6 +206,14 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 						.dataManager="${this.ouFilterDataManager}"
 						@d2l-labs-ou-filter-change="${this._orgUnitFilterChange}"
 					></d2l-labs-ou-filter>
+					<d2l-input-date-range
+						max-value="${this._maxValue}"
+						min-value = "${this._minValue}"
+						end-value = "${today}"
+						start-value = "${twoWeeksAgo}"
+						class = "d2l-insights-date-range-offset"
```
Vitalii-Misechko, line 214: Why do you need this class? I don't see any CSS linked to it

pages/adoption-dashboard-view.js
```diff
@@ -273,6 +294,25 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		const upperDateBound = new Date(startVal);
+
+		lowerDateBound.setMonth(lowerDateBound.getMonth() - 6);
+		lowerDateBound.setDate(lowerDateBound.getDate() + 1); //+1 for today inclusive
+		upperDateBound.setMonth(upperDateBound.getMonth() + 6);
+		upperDateBound.setDate(upperDateBound.getDate() + 1);
+
+		this._minValue = formatDate(lowerDateBound, { format: 'yyyy-MM-dd' });
+		this._maxValue = formatDate(upperDateBound, { format: 'yyyy-MM-dd' });
+		event.stopPropagation();
```
Vitalii-Misechko, line 311: AFAIK `stopPropagation` position in the method does not change its behavior. Therefore, I would place it at the first line in order other developers to find it immediately when they open the method. Do we need this line at all?

Vitalii-Misechko (commented):
Looking good! Added a few thoughts

I don't find these spacings good. I hope the new designer will suggest something.

image

pages/adoption-dashboard-view.js
Sean-Kurian: Removed it, I had initially linked CSS to it until Nick pointed out that I could use the tag directly instead of the class.

pages/adoption-dashboard-view.js
Vitalii-Misechko: I would remove it and double-check if there are no issues.

pages/adoption-dashboard-view.js
```diff
@@ -313,6 +354,28 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		lowerDateBound.setDate(lowerDateBound.getDate() + 1); //+1 for today inclusive
+		upperDateBound.setMonth(upperDateBound.getMonth() + 6);
+		upperDateBound.setDate(upperDateBound.getDate() + 1);
+
+		this._minValue = formatDate(lowerDateBound, { format: 'yyyy-MM-dd' });
+		this._maxValue = formatDate(upperDateBound, { format: 'yyyy-MM-dd' });
+
+		const startDay = new Date(startVal);
+		const endDay = todayDateString.toDateString() === new Date().toDateString() ? null : new Date(endVal);
+		this.data.load(this, { startDay: startDay, endDay: endDay });
```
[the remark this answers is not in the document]
Sean-Kurian: This should be fixed, was unsure if I needed to make any further changes to anything you wrote but it seems to work

johngwilkinson:
Tested various combinations of dates and drill e2e.
Confirmed today works (sends null), and time ranges beginning more than six months ago work (provided the end date is moved back first).

Validation bug:
- set end date to a week ago (for example)
- set start date to earliest possible date
- now open end date picker and click "Today"
  - expected: no effect/validation error but no load
  - actual: more than six months of data loads; on losing focus, end date turns red for invalid range
- set start date up so that end date is valid
  - expected: validation red highlight disappears from end date
  - actual: end date remains red; workaround is to select a different end date then today's date again

johngwilkinson:
In another PR: should set up `syncWithUrl` on a String-valued getter/setter for data._dateRange, so that date is included in browser history and bookmarks

locales/en.js
```diff
@@ -17,7 +18,7 @@ export default {
 	"dashboard:rootOrgUnit": "All",
 
 	"courseAccessCountsCard:courseAccess": "Course Access Counts",
-	"courseAccessCountsCard:numberOfUsers": "Total for last 2 weeks",
+	"courseAccessCountsCard:numberOfUsers": "Total for Time Period",
```
bemailloux, line 21: The Tool Usage Comparison card x-axis label might need updating too. Unless you're applying changes there in a different PR?

bemailloux (commented):
Looks like the mobile skeleton visual-diffs show the loaded page instead of skeletons. Fixing can probably be deferred to a different PR but it should happen.

Otherwise LGTM - going to do some testing before approval

bemailloux:
It looks like if the back-end's default date range doesn't match the front end's default range, the ranges can be out of sync on first load
image

johngwilkinson:
> It looks like if the back-end's default date range doesn't match the front end's default range, the ranges can be out of sync on first load

Agreed, although this isn't a defect so long as we make them match. So it's more of a place where the code could be more careful - but I'm not sure if it's worth it in this case

model/data.js
```diff
@@ -59,6 +56,17 @@ export class Data {
...
 	}
 
+	getDateRange() {
+		return this._dateRange;
+	}
+
+	setDateRange(dateRange) {
+		if (typeof this._dateRange === 'undefined' || this._dateRange.startDay !== dateRange.startDay || this.dateRange.endDay !== dateRange.endDay) {
+			this._dateRange = { startDay: dateRange.startDay, endDay: dateRange.endDay };
+			this.load({});
```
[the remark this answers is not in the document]
Sean-Kurian: I don't believe so, unless there's a timezone difference in the initial (but I think John mentioned that his idea of fixing by sending null should take care of that)

bemailloux:
Retest: found one other strange thing. If you set the start date to 2022-02-27, then to 2022-03-30, Tool Usage Trend and Tool Usage by Role graphs have the wrong start date (2022-02-27).

Sean-Kurian:
This should conclude all bugs to be fixed for this PR. Going to open a new PR to fix the validation bug John found, but I believe all errors for this PR have been resolved. 

Issues found by Bryan: 
update tool chart label (RESOLVED)
localize date range (RESOLVED)
alert.localize is not a function (RESOLVED) 
end date can be set beyond today (RESOLVED)

Inconsistent Tool Usage Trend and Tool Usage By Role Graphs (in progress)

Unresolved (in this PR) errors: 
front-end date range is not applied on first server load
visual-diffs are date dependent
mobile skeleton visual diffs show fully loaded page
validation - syncWithUrl bug

bemailloux:
Not sure if you saw this additional issue I found:

> Retest: found one other strange thing. If you set the start date to 2022-02-27, then to 2022-03-30, Tool Usage Trend and Tool Usage by Role graphs have the wrong start date (2022-02-27).

In other words, if you set the start date to some value where there is no tool usage data for the period, the resulting start date is incorrect. I think this can also be resolved by applying the start date after a load, so this can probably be deferred

locales/en.js
```diff
@@ -11,6 +11,7 @@ export default {
 	"dashboard:EnrollmentsAndWithdrawalsSection": "Enrollments and Withdrawals",
 	"dashboard:learnMore": "Learn More",
 	"dashboard:AllError": "Data for this dashboard could not be retrieved. Please contact D2L Support if this problem persists.",
+	"dashboard:DateRangeFilter": "Date Ranger Filter",
```
bemailloux, line 14: `Ranger` 🤠

bemailloux (approved):
LGTM!

insights-adoption-dashboard.js
```diff
@@ -130,7 +130,7 @@ class AdoptionDashboard extends Localizer(MobxLitElement) {
 				dataSetFactories,
 				dataApiClient: new DataApiClient(this.metronEndpoint)
 			});
-			this._data.load({ alert: this });
+			this._data.load(this, undefined);
```
johngwilkinson, line 133: minor, maybe clean up in the next PR: `data.load()` only takes one parameter now

model/data.js
```diff
@@ -59,6 +56,17 @@ export class Data {
 		return this._resultSets.some(r => r.isQueryError);
 	}
 
+	getDateRange() {
```
johngwilkinson, line 59: nit: it's just a stylistic difference, but this could be a getter and setDateRange could be a setter (`get dateRange(){...}`) - no change needed

johngwilkinson (approved):

bemailloux (approved):

model/data.js
Sean-Kurian: Missed that there was a difference, but is this true for setDateRange? I need to pass two parameters in for it and lint only allows one.

johngwilkinson: A setter takes only one parameter, which should generally be the same kind of object the getter returns. So that `data.dateRange = data.dateRange` causes no errors and has no effect.

PR #95, opened 2022-03-15
US135382: ou drill via course access card

johngwilkinson:
Quality Notes
- functional
  - e2e w. server changes on draco
  - drill down, use browser back button to go back up
  - caching works (based on timing)
  - also used browser forward button to re-drill
  - charts all appear to have correct data
  - labels only appear as links and are only clickable if they are not course offerings
  - page reload on a drill loads the drill (i.e. url can be bookmarked)
- perf
  - server load on each drill, including drilling back up - could consider http caching to help with this, but it's not too bad as-is
  - other than that, no change
- security/devops: n/a
- ux/docs
  - link-styled labels were approved in demo

image


model/urlParameter.js
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	set value(v) {
+		this._object[this._field] = v;
+	}
+}
+
+export function syncWithUrl(object, field, key) {
+	// ContextReactor will hold on to a reference to this object
+	new UrlParameterDecorator(object, field, key);
```
Vitalii-Misechko, line 108: I like how `UrlParameterDecorator` links url parameter to an object without making object more complex

pages/adoption-dashboard-view.js
```diff
@@ -251,6 +252,11 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 		`;
 	}
 
+	_drill(event) {
+		event.stopPropagation();
+		this.data.parentId = event?.detail?.orgUnitId;
```
Vitalii-Misechko, line 257: I assume that it's okay when `parentId` is assigned to `undefined` value even when this case is not handled explicitly in the setter of `parnetId`

components/cards/course-access-counts-card.js
```diff
@@ -47,9 +55,17 @@ class CourseAccessCountsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 				.options="${this._chartOptions}"
 				.globalOptions="${this._globalOptions}"
 				?skeleton="${this.dataSet.isLoading}"
+				.axisClickHandler="${this.drill}"
```
bemailloux, line 58: Wouldn't we normally pass an event back up from the chart, instead of passing a handler down into the chart for it to execute?

(Although I'm not sure if it makes much of a difference beyond vaguely adhering to the "separate display / business logic" idea here.)

model/dataSets/toolUsageTrendByToolDataSet.js
```diff
@@ -10,12 +10,16 @@ export class ToolUsageTrendByToolDataSet extends DataSet {
...
+		// *unless* we are drilling, which case we show all
 		const rows = data?.toolUsageTrendData ?? [];
-		const rootOrgUnitId = data?.lmsData?.rootOrgUnitId; // if this isn't provided, no data will appear
 
-		const rowsForRootOrgUnit = rows.filter(r => r.o === rootOrgUnitId);
+		let rowsToAggregate = rows;
+		if (data?.lmsData?.isParentRoot) {
+			const rootOrgUnitId = data?.lmsData?.rootOrgUnitId; // if this isn't provided, no data will appear
+			rowsToAggregate = rows.filter(r => r.o === rootOrgUnitId);
+		}
```
Vitalii-Misechko, line 20: Should we improve unit tests for this class?

model/dataSets/toolUsageTrendByRoleDataSet.js
```diff
@@ -10,12 +10,16 @@ export class ToolUsageTrendByRoleDataSet extends DataSet {
 		 */
 	process(data) {
 		// rows now includes data for root org unit and all its children. For this card we only want to show root data
+		// *unless* we are drilling, which case we show all
```
bemailloux, line 13: 👍

pages/adoption-dashboard-view.js
johngwilkinson: Yes, `undefined` is the default anyway

bemailloux, line 257: At the moment, the only way to get back to a parent org unit is by manually setting the URL, right?

Vitalii-Misechko (approved):
LGTM!

model/urlParameter.js
```diff
@@ -0,0 +1,109 @@
+import { ContextReactor, redirect } from '@brightspace-ui-labs/lit-router';
+import { autorun } from 'mobx';
+
+export class UrlParameter extends ContextReactor {
```
bemailloux, line 4: Is this class testable at all?

johngwilkinson: This comes right from AQD (where it's called `FilterState`), so I'll see if there's anything there

pages/adoption-dashboard-view.js
johngwilkinson: yes/browser back button. Breadcrumbs are part of the story, coming soon :)

components/cards/course-access-counts-card.js
johngwilkinson: I had that thought too, but everything else in Highcharts options works in a "pass in a function" kind of way, so I stuck with that here

model/urlParameter.js
johngwilkinson: There don't seem to be any tests for this class in AQD. :( It should probably move to brightspace-ui-labs at some point (dragonslayer?) - maybe testing can happen then

bemailloux (dismissed):

bemailloux (approved):

Vitalii-Misechko (approved):
LGMT!

PR #108, opened 2022-03-21
replace csv-stringify with an esm-compatible lib

johngwilkinson:
There is a discrepancy between dev and prod BSI such that not all package.json `exports` syntax is supported in prod. There's no easy workaround for this (dynamic imports might work, but likely would not prevent an error during rollups), so this PR switches back from csv-stringify to export-from-json. The latter has a more awkward syntax resulting in some extra processing, but the end result is the same.

Quality Notes
- functional: e2e and demo page on draco stack
  - dev mode
  - prod mode test needs to wait until this version is in BSI (npm link doesn't work w. prod BSI builds)
- perf: if the data size is very large the extra processing will slow things down and use up extra memory
- security/devops/ux/docs: n/a


bemailloux:
I thought the reason we replaced export-from-json in the first place was because there was some issue with zipping?

johngwilkinson:
Oh, btw: I searched pretty exhaustively for another browser-friendly, esm-export lib to covert to csv, and I couldn't find any. Tons of csv converter libs, but mostly node oriented and not browser-compatible. The one that should have been ideal is papaparse, but they don't have an es6/esm module version and I couldn't get it to import without errors.

model/exportData.js
```diff
@@ -21,11 +21,25 @@ function saveAs(content, fileName) {
...
 	const zipData = {};
 	dataSets.forEach(d => {
 		const fileName = `${localize(d.exportFilenameTerm)}.csv`;
-		const csv = stringify([d.headersForExport.map(localize), ...d.dataForExport]);
+		const headers = d.headersForExport.map(localize);
+		const data = toObjects(d.dataForExport, headers);
+		const csv = exportFromJSON({
+			data,
+			exportType: 'csv',
+			processor: csv => csv
```
bemailloux, line 41: Do we need any special options to handle CSV edge-cases? Like double-quotes to escape fields with commas in them, double-double-quotes to escape double-quotes, ...

bemailloux (approved):
LGTM!

package.json
```diff
@@ -64,9 +64,9 @@
     "@brightspace-ui/intl": "^3.2.0",
     "array-flat-polyfill": "^1.0.1",
     "chai-a11y-axe": "^1.3.2",
-    "csv-stringify": "^6.0.5",
```
Vitalii-Misechko, line 67: Could you also remove `csv-stringify` from `karma.conf.js` and `karma.sauce.conf.js`? You just need to delete `plugins`

Vitalii-Misechko (approved):
LGTM!

package.json
johngwilkinson: Done, thanks

model/exportData.js
johngwilkinson: Nope: did a test with a role called
```
`Instructor, 
		test"csv"`
```
in server.js and it was properly escaped (the excel cells contain the whole string including carriage return, tabs, quote, and comma

bemailloux (approved):

Vitalii-Misechko (approved):

PR #109, opened 2022-03-22
Us135382/drill tweaks

johngwilkinson:
Quality Notes
- functional
  - just tested the demo page due to current lit 2 migration
  - charts load as expected
  - tool usage and course access now match (tooltips, truncation, ids, titles, sorting)
    - removed the x-axis title from tool usage to match course access and design
  - drill via mouse works as expected (no new data in the demo, but the url updates and the page reloads)
  - drill via keyboard with and without Narrator enabled
    - focus skips course offering (not drillable)
- perf: n/a
- security: n/a
- devops: n/a
- ux/docs
  - screenreader experience seems consistent with other charts
  - I put the axis labels after the series in tab order - this seemed to me to work best, as likely the user wants to check the data before drilling

image


components/chart/chart.js
```diff
@@ -239,9 +245,25 @@ class Chart extends SkeletonMixin(Localizer(LitElement)) {
...
+				el.addEventListener('click', () => this.axisClickHandler(el.dataset.id));
+				customComponents[`custom_${el.dataset.id}`] = new this.AccessibleClickable(el, this.chart);
+			});
+
+		// tell Highcharts about the new accessibility components and what order to navigate them in
+		this.chart.update({
+			accessibility: {
+				customComponents,
+				keyboardNavigation: {
+					order: [ 'container', 'series', ...Object.keys(customComponents), 'zoom', 'rangeSelector', 'chartMenu', 'legend' ],
```
NicholasHallman, line 263: Do we need `container` first for the axis selectable chart? Feels inconsistent with the other charts where tabbing in selects the data.

NicholasHallman: `nit` On that note for `...Object.keys(customComponents)` It would be nice if tabbing selected the first element and you could use the arrow keys to navigate, tabbing again would bring you to the next chart. Similar to thedata in the other charts.

johngwilkinson: It is inconsistent, but I like it :) Maybe we should think about adding it to the others too. I find when using a screen-reader it's confusing to land directly on the first series with no context about which chart it is.

johngwilkinson: About tabbing and arrow keys: it does work that way in some directions but not others. We'd have to somehow override how the built-in components like series work to have arrow keys go through the whole thing bottom-to-top, but top-to-bottom works fairly well.

NicholasHallman: > I find when using a screen-reader it's confusing to land directly on the first series with no context about which chart it is.

Weirdly enough the screen reader does give context but only for system access. It doesn't read out the chart names of the other charts on main and the accessibility debugger shows they are the same so not sure what's going on there

NicholasHallman: And for tabbing, its not that inconsistent when thelegend is taken into account so I'm cool with leaving it :)

NicholasHallman:
`nit` For some reason Silly walks doesn't say `drill into silly walks` when using narrator

components/chart/chart.js
Vitalii-Misechko: > > I find when using a screen-reader it's confusing to land directly on the first series with no context about which chart it is.
> 
> Weirdly enough the screen reader does give context but only for system access. It doesn't read out the chart names of the other charts on main and the accessibility debugger shows they are the same so not sure what's going on there

I would add that this depends on a screen reader. NVDA provides context and the Narrator doesn't. I like @johngwilkinson's solution because it is consistent among NVDA and Narrator

components/chart/accessibleClickable.js
```diff
@@ -0,0 +1,49 @@
...
+	const AccessibleClickable = function(element, chart) {
+		this.initBase(chart);
+		this.chart = chart;
+		this.element = element;
+		this.name = element.textContent;
+	};
+
+	AccessibleClickable.prototype = new Highcharts.AccessibilityComponent();
+
+	Highcharts.extend(AccessibleClickable.prototype, {
```
Vitalii-Misechko, line 11: `curious` I assume we just can't use `extends` syntax, right?
```JS
class AccessibleClickable extends Highcharts.AccessibilityComponent {
  constructor (element, chart) {}
  onChartUpdate() {}
  getKeyboardNavigation() {}
}
```

model/dataSets/courseAccessDataSet.js
```diff
@@ -1,17 +1,30 @@
+import { compareNameAndId } from '../util';
 import { DataSet } from '../dataSet';
 
+function sorted(data) {
+	const orgUnits = data.courseAccessData?.orgUnits ?? [];
+	// sort the org units by name and id, then use the resulting indices to order each series to match
+	const sortIndices = orgUnits.map((o, i) => i).sort((a, b) => compareNameAndId(orgUnits[a], orgUnits[b]));
```
Vitalii-Misechko, line 7: neat!

model/dataSets/toolUsageComparisonDataSet.js
```diff
@@ -55,12 +56,10 @@ export class ToolUsageComparisonDataSet extends DataSet {
 			, [ 'name', 'id' ] // same sorting as in Tool Usage Trend by Tool
```
Vitalii-Misechko, line 56: we can replace `sortBy` with `sort(compareNameAndId)` right?

Vitalii-Misechko (approved):
LGTM!

model/dataSets/toolUsageComparisonDataSet.js
johngwilkinson: I believe so

components/chart/accessibleClickable.js
johngwilkinson: I'm roughly following Highcharts' example: https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/accessibility/custom-component

I haven't tried it that way but I'm assuming it won't work because `AccessibilityComponent` isn't declared using `class`

johngwilkinson:
> `nit` For some reason Silly walks doesn't say `drill into silly walks` when using narrator

This doesn't match my testing - what does it say for you? One thing I did see sometimes was occasionally it would skip a label altogether if you hit tab at the "wrong" time - but I didn't see that with this particular ordering.

NicholasHallman (approved):

PR #111, opened 2022-03-22
US135382: avoid extra history entry on first load

johngwilkinson:
This fixes a (not-yet-shipped) bug similar to one in AQD, where the initial page load rewrites the url resulting in an extra browser history entry, so the user has to click "back" twice to get back to the Insights Portal.

quality notes: just demo page testing


model/urlParameter.js
```diff
@@ -67,8 +70,7 @@ export class UrlParameter extends ContextReactor {
 		const currentSearchValue = searchParams.get(this.key);
 		// We don't need to push a new url state if nothing has changed from the last save
 		// This also keeps us from updating the state when the back button is pressed.
-		if (currentSearchValue === this.value ||
-			(currentSearchValue === undefined && (this.value === '' || this.value === undefined))) {
+		if (currentSearchValue === this.value || !currentSearchValue && !this.value) {
```
johngwilkinson, line 73: matches `null` and `''`

model/urlParameter.js
```diff
@@ -39,7 +39,10 @@ export class UrlParameter extends ContextReactor {
 
 		clearTimeout(this._activeDebounce);
 		this._activeDebounce = setTimeout(() => {
-			const url = `${location.pathname}?${this._paramsDebounce.toString()}${location.hash}`;
+			const queryString = this._paramsDebounce.toString();
+			const url = queryString ?
+				`${location.pathname}?${queryString}${location.hash}` :
+				`${location.pathname}${location.hash}`;
```
NicholasHallman, line 45: So simple! Nice catch

NicholasHallman (approved):

Vitalii-Misechko (approved):

PR #113, opened 2022-03-24
Bump version to match release tags in Github

annabelsegalld2l:
Looks like BSI is having some issues picking up the adoption-dashboard releases from Github. I think this is because the release tags and version in the `package.json` aren't matching up, and this became an issue after the major version bump. I will make a manual release in Github for 2.0.3.

If you are interested, you can also add semantic releasing to your repo by adding this file (no need to change anything) https://github.com/BrightspaceUI/create/blob/main/.github/workflows/release.yml.


johngwilkinson (approved):
Thanks! Just a reminder I noticed the ref in BSI's package.json also needed to be bumped from `^1` to `^2`.

annabelsegalld2l:
> Thanks! Just a reminder I noticed the ref in BSI's package.json also needed to be bumped from `^1` to `^2`.

Yep, it already is: https://github.com/Brightspace/brightspace-integration/blob/main/package.json#L50. I'm having issues where any more `npm install` commands in BSI are deleting adoption-dashboard from the package-lock and I think the version mismatch here is the reason.

Vitalii-Misechko (approved):

PR #114, opened 2022-03-24
US132923 [Adoption][Org Filter] Apply Tree Filter Selection API

Vitalii-Misechko:

API defined here: https://github.com/Brightspace/bdp-query/pull/201

Outside the scope:
* Tree search
* Application of tree selection

### Quality Notes

* Functional Testing 
  * [x] E2E Test cases
    * [x] Tree filter shows departments and courses
    * [x] Departments can be opened and dynamicly loaded
    * [x] Filter will display default selection (non is selected)
    * [x]  User will have the ability to open filter, deselect or maintain defaults and select new filter options
      * [x] Once cleared, no filters will be applied and all course data associated to the user will be shown (**separate US**)
    * [x] User will have the ability to search through filter results (**separate US**)
    * [x] Filter results should be displayed alphabetically
* Security, devops, perf: N/A
* UX and docs: N/A

* [ ] Demo

#### Screenshots

image


model/ouFilter.js
```diff
@@ -54,24 +49,21 @@ export class AdoptionOuFilterDataManager extends OuFilterDataManager {
...
 		const lastSearchResults = this._apiClient.fetchLastSearch();
 
 		this._orgUnitTree = new Tree({
 			// add in any nodes from the most recent search; otherwise
 			// the search will blink out and come back, and also drop any "load more" results
 			nodes: lastSearchResults ? [...this._apiData.OrgUnits, ...lastSearchResults] : this._apiData.OrgUnits,
 			leafTypes: [COURSE_OFFERING],
-			invisibleTypes: [this._apiData.SemesterTypeId],
-			selectedIds: this._apiData.SelectedOrgUnitIds || [],
+			invisibleTypes: [],
```
johngwilkinson, line 59: we should probably hide semesters (follow-up PR perhaps, since we don't have the semester type id handy)

model/data.js
```diff
@@ -45,6 +46,17 @@ export class Data {
 		this.load({});
 	}
 
+	get orgUnitIds() {
+		return this._orgUnitIds ?? [];
```
johngwilkinson, line 50: Can you do a `syncWithUrl` on `orgUnitIds`? I think it's fine that it's an array, but if you need an additional layer of conversion, you can do that too, like in this PR I'm working on: https://github.com/Brightspace/insights-adoption-dashboard/pull/115/files#diff-eee75df07aac258797ab7469564820082ce3f54d26a380d5998fed4a5bf130d4

model/ouFilter.js
Vitalii-Misechko: I see that `GetVisibleChildren` filters semesters already (see picture). I guess we can omit semester type here
image

model/data.js
Vitalii-Misechko: Why do we want to persist `orgUnits` in URL? Did you mean `selectedOrgUnits`? If yes, what is the default selected list (all)?

johngwilkinson: I was assuming this was the selected org units - maybe I misread. But yes, I think selected org units should be in the URL so that the view can be bookmarked, refreshed, shared, and responds to the browser back and forward buttons.

The default would be no selection, which means "show all children of the currently drilled node (root by default)"

model/ouFilter.js
johngwilkinson: True!

model/data.js
Vitalii-Misechko: I removed `orgUnitIds` property

Vitalii-Misechko:
> Filter results should be displayed alphabetically

Is this AC about search results?

johngwilkinson:
> > Filter results should be displayed alphabetically
> 
> Is this AC about search results?

I'm not sure, but I think it may mean that, but also that the children of each org unit should be sorted alphabetically in the tree.

model/data.js
johngwilkinson: Just looking at the story A.C., and I wonder if the idea is that if there are more than 100 children of the current org unit (root or parentid), so we're only showing 100 of them, the org unit tree filter should automatically select the ones in view. Engagement works that way, but it defaults to only 10 courses, I believe - so I'm not sure we should do the same thing here, since that's a lot of boxes to uncheck. It's probably easier to use if the user can just go in and check a few things they actually want to see. Might need to get a PM/design decision on that.

model/ouFilter.js
```diff
@@ -50,36 +55,55 @@ export class AdoptionOuFilterDataManager extends OuFilterDataManager {
...
-		return this._orgUnitResultSet.data.OrgUnits ? this._orgUnitResultSet.data : defaultApiData;
+		return this._orgUnitResultSet.data.orgUnits ? this._orgUnitResultSet.data : defaultApiData;
+	}
+
+	get _treeOrgUnits() {
+		const orgUnits = this._apiData.orgUnits;
+
+		// In order to render the tree
+		// these nodes should include at least one root node that has 0 in Parents array
+		if (orgUnits.length > 0 && !orgUnits.some(o => o.Parents.includes(0))) {
```
johngwilkinson, line 66: Seems like this shouldn't be necessary based on what I see in the bdp-query PR? I guess it's not bad to have it though.

Vitalii-Misechko:
> > > Filter results should be displayed alphabetically
> > 
> > 
> > Is this AC about search results?
> 
> I'm not sure, but I think it may mean that, but also that the children of each org unit should be sorted alphabetically in the tree.

In this case, there is nothing to do. The component has built-in sorting. For search, we need to sort and limit results at the server-side I think

johngwilkinson (approved):

model/ouFilter.js
Vitalii-Misechko: LMS response doesn't include a root node. But, I remember talks that we might want to return it (bdp-query code assume that we can return a root node in the future).
ou-filter doesn't work without a root node. So, this code ensures that UI will work regardless presence of a root node in the API response

model/data.js
Vitalii-Misechko: @johngwilkinson 

>  I believe - so I'm not sure we should do the same thing here, since that's a lot of boxes to uncheck. 

Is that a performance concern? 100 courses don't seem many.

> It's probably easier to use if the user can just go in and check a few things they actually want to see. 

You mean that ou-filter doesn't have selected nodes at first load, right? So, only user action changes selected nodes

model/ouFilterApiClient.js
```diff
@@ -23,21 +33,24 @@ const orgUnitSearchCache = {
...
+
+		if (!Array.isArray(resObj)) {
+			throw new Error('Can`t handle result that is not array');
 		}
 
-		if (!relevantChildrenCache.has(orgUnitId)) {
-			relevantChildrenCache.set(orgUnitId, results);
+		const results = {
+			Items: mapVisibleChildren(resObj, orgUnitId),
+			PagingInfo: {} // API doesn't support paging
```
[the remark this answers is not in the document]
Vitalii-Misechko: >As mentioned elsewhere, I believe this will need to change (fine for this PR though)

`PagingInfo` is a part of the API that is hardcoded at `ou-filter` level. Not sure what we can do about it.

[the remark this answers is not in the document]
johngwilkinson: I mean we need to add paging to that API

model/ouFilterApiClient.js
```diff
@@ -23,21 +33,24 @@ const orgUnitSearchCache = {
...
+		};
+
+		if (!visibleChildrenCache.has(orgUnitId)) {
+			visibleChildrenCache.set(orgUnitId, results);
```
bemailloux, line 51: If an org unit has multiple parents, does this mean it will only appear under one of its parents in the tree (instead of both of its parents)?

bemailloux (approved):

model/ouFilterApiClient.js
Vitalii-Misechko: The short answer is I don't know. This API allows the node to appear for both parents. I also assume that two parents' case is available when the node is under Semester and Department. Because this tree doesn't show semesters then I hope we will not observe such cases. Let me know if you want more info so I dig into ou-filter implementation

bemailloux: It's possible in the LMS to add a course under multiple Departments. It's not common but since there's nothing stopping that, we have to assume that some clients have that kind of setup and so we need to handle it. But if the API allows the node to appear for both parents then it sounds like you've handled it already 👍

PR #115, opened 2022-03-24
US135382: breadcrumbs

johngwilkinson:
Note: I expect to add some unit tests in another PR

Solution overview: 
- stores the full org unit drill path in the url for bookmarking, sharing, reloading, and navigation
  - names and ids
  - can't rely on server to determine drill path because it's not uniquely determined by the parent node
  - storing names saves extra server lookups
  - names are base64 encoded in case they contain commas (list delimiters)
  - this results in big obscure urls, but well below any technical limits
- breadcrumbs don't wrap; this design simply allows the card to grow
  - alternatives are to have a scroll-bar or to cut off the breadcrumbs
    - cutting off is a usability problem
    - the scroll bar appears before it is really needed and looks bad (shifts chart down)
    - growing card doesn't look great, but for many orgs it won't happen
    - when the chart grows enough, it moves down below the System Access chart
  - for display, breadcrumbs are truncated to 15 characters: allows for sufficient levels of drill without expanding the card
    - but on the url, they are allowed to be longer (still with a limit), and title (hover text) and aria-label use the longer name
    - because the breadcrumb component does not support css truncation without also hiding the trailing arrow symbol, truncation is at a character count, not a pixel count, and there are three periods instead of a proper ellipsis

Quality Notes
- functional (demo page and e2e on draco2)
  - drill down and use breadcrumb to go back up one level or multiple levels at once
  - navigate forward and back using browser buttons
  - reload page after drill
- perf: n/a
- security
  - I don't think it's a significant concern, but it's possible to construct urls that display the wrong names for org units - but the data shown and accessible is not affected
- devops: n/a
- ux/docs:
  - see notes above
  - see visual diffs for screenshots
  - [ ] demo


Vitalii-Misechko:
The first org unit is not clickable. Is this expected?

image

demo/ad-demo-view.js
```diff
@@ -13,13 +13,14 @@ class AdDemoView extends LitElement {
 		let serviceWorker;
 		if (hashMap.has('immediate')) {
 			serviceWorker = worker;
+		} else {
+			if (hashMap.has('delay')) {
+				const [initial, result] = hashMap.get('delay').split(',');
+				initialDelayMs = parseInt(initial) || initialDelayMs;
+				resultsDelayMs = parseInt(result) || resultsDelayMs;
+			}
+			serviceWorker = getWaitingWorker({ initialDelayMs, resultsDelayMs });
```
Vitalii-Misechko, line 22: Thank you for fixing it

johngwilkinson:
> The first org unit is not clickable. Is this expected?
> 
Yes, that one is type 3 (course offering) in the demo data; courses can't be drilled into

bemailloux:
> > The first org unit is not clickable. Is this expected?
> 
> Yes, that one is type 3 (course offering) in the demo data; courses can't be drilled into

I feel like we might need to make that clear somewhere. It seems like users often expect to be able to drill down to the section level (or at least, that seems to be a common ask).

johngwilkinson:
> > > The first org unit is not clickable. Is this expected?
> > 
> > 
> > Yes, that one is type 3 (course offering) in the demo data; courses can't be drilled into
> 
> I feel like we might need to make that clear somewhere. It seems like users often expect to be able to drill down to the section level (or at least, that seems to be a common ask).

It's in the A.C. and matches the current Adoption Dashboard, so I assume it's also documented. Having a reminder in the UI is an interesting idea - tooltip maybe - but I'm not sure where to put it that wouldn't be annoying for screenreader users.

johngwilkinson:
> > names are base64 encoded in case they contain commas (list delimiters)
> 
> "/" and "+" are valid b64 characters, but probably shouldn't be allowed in URLs

Thanks for the note. Values get uriencoded when written to the URL, so that's not an issue. The encoding here is just to get rid of commas, and base64 was a quick way to handle it. (I tried uri encoding, but it got automatically double-decoded, resulting in an extra update and history entry -- so base64 is a way of staying away from doubling up on uri encoding)

Vitalii-Misechko (approved):
LGTM!

bemailloux (approved):
LGTM!

bemailloux (approved):

Vitalii-Misechko (approved):

PR #122, opened 2022-03-28
Fix error `TypeError: _e(...).groupBy is not a function`

Vitalii-Misechko:
Support tree shaking with `lodash-es`. Fix error `TypeError: _e(...).groupBy is not a function`.

The fix is based on assumption that `lodash-es` fails to load its `groupBy` properly. Therefore, we specify `groupBy` explicitly. As a side effect, the code loads only necessary modules of this library.

* E2E with local BSI continues to work without errors


johngwilkinson:
Did you test with BSI in prod mode?

johngwilkinson (approved):

PR #128, opened 2022-03-29
Us135382/breadcrumb tests

johngwilkinson:
Tests, plus some UX changes requested by Design.
1. Appearance when there are more than 4 levels of drill: 
image

2. Leave a gap instead of the word "All" when there's no drill: 
image

Note that this is somewhat temporary: the plan is to move the breadcrumbs up top with the filters, but that design isn't yet ready. For this reason, I haven't added tests for the above two behaviours.

Quality: tested case (1) using demo page; tested both cases/smoke on draco


bemailloux:
In the "All > ... > ... > ... > ... > {current}" case: 
(1) clicking on a "..." will bring you to that drill level, right?
(2) can the user figure out what org unit each "..." corresponds to somehow? e.g. by hovering over the link?

johngwilkinson:
> In the "All > ... > ... > ... > ... > {current}" case: (1) clicking on a "..." will bring you to that drill level, right? (2) can the user figure out what org unit each "..." corresponds to somehow? e.g. by hovering over the link?

Oh, good question! It does work that way, yes. And I thought the labels would work, but I forgot to check, and in fact I have a bug where there's no title on the ellipsis crumbs. Fix incoming...

bemailloux (dismissed):
LGTM!

components/cards/course-access-counts-card.js
```diff
@@ -85,8 +90,9 @@ class CourseAccessCountsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 
 		return html`<d2l-breadcrumbs class="d2l-card-breadcrumbs">
 			${crumbs.map((d, i) => {
-		const name = d.name || rootName;
-		const shortName = this._shorten(name, maxDrillNameLength);
+		const isRoot = !d.name;
+		const name = isRoot ? rootName : d.name;
+		const shortName = (isLong && !isRoot) ? '...' : this._shorten(name, maxDrillNameLength);
 		return html`<d2l-breadcrumb
```
Vitalii-Misechko, line 96: `thoughts on code design, no changes needed` I assume that we can pull this logic inside of `d2l-breadcrumb` sometimes in the future

Vitalii-Misechko (approved):
LGTM! Good tests! 👍

bemailloux (approved):

PR #134, opened 2022-03-29
Us135383/tool comparison drill

johngwilkinson:
Add drill and breadcrumb to Tool Usage Comparison by refactoring so that the two data-by-ou-with-drill cards use the same component.

Quality Notes
- functional
  - demo: drill far enough to get "..." breadcrumbs
  - demo/e2e: drill up and down using a combination of the two cards and the browser history buttons
- security: n/a
- perf: n/a
- devops: depends on bdp-query change https://github.com/Brightspace/bdp-query/pull/204
- ux/docs
  - same as course access
  - demoed


components/cards/bar-chart-with-drill-card.js
```diff
@@ -18,44 +18,51 @@ const MIN_SIZE_PER_ROW = 13; // px
 const maxDrillNameLength = 15;
 const chartHeight = CHART_HEIGHT - 21; // exclude height of the breadcrumbs
 
-class CourseAccessCountsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
+class BarChartWithDrillCard extends SkeletonMixin(Localizer(MobxLitElement)) {
```
bemailloux, line 21: Nice!

components/cards/bar-chart-with-drill-card.js
```diff
@@ -18,44 +18,51 @@ const MIN_SIZE_PER_ROW = 13; // px
...
 	get _cardCategoryObjects() {
 		return this.data ?
-			this.data.orgUnits
+			this.data.categories
 				.map(o => ({
 					name: this.localize('chart:labelWithId', o),
 					ariaLabel: this.localize('chart:drillIntoLabelWithId', o),
 					id: o.id,
-					clickable: o.orgUnitTypeId !== courseOfferingTypeId
+					clickable: Number(o.typeId) !== this.stopType
```
bemailloux, line 65: Is the drill behaviour for ToolUsageComparison fully defined yet? I know Binu had to stop cause they took it back to the drawing board, and I'm wondering if "stopType" will make sense in the TUC context.

bemailloux (approved):
LGTM! Love the deduplication work you put in here

components/cards/bar-chart-with-drill-card.js
johngwilkinson: Yes, I believe this is everything for TUC. It was Tool Usage Trend that got changed - it was going to be drill by tool then drill by ou I think, but now I don't think that one will drill at all

components/cards/bar-chart-with-drill-card.js
```diff
@@ -64,9 +71,8 @@ class CourseAccessCountsCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 		return html`
 			<div class="d2l-card-title d2l-skeletize d2l-skeletize-45 d2l-body-standard">${this.title}</div>
 			${this._breadcrumbs}
-			<d2l-labs-chart
+			<d2l-labs-chart class="${ifDefined(this.isDoubleWidth ? 'd2l-insights-summary-card-double-width' : undefined)}"
```
Vitalii-Misechko, line 74: `nit` I believe lit-element devs created `classMap` for cases like this

components/cards/bar-chart-with-drill-card.js
Vitalii-Misechko: 🚀

PR #142, opened 2022-04-01
US135390: Adoption: role filtering via chart

johngwilkinson:
Server side: https://github.com/Brightspace/bdp-query/pull/206

UPDATE: now includes filter interactions: clicking on bars in Course Access or areas in Tool Usage by Role will toggle filtering
by role.

This is just a sort of full-stack refactor: no change to behaviour yet, but System Access and Enrollments&Withdrawals now come from the server broken out by role so they can, next, be filtered. I also aligned then with tool usage in that the server does not fill in days with no data.

- [ ] reviewer 1 tested
- [x] reviewer 2 tested

Quality Notes
- functional: e2e & demo, including csv export (retested w. latest changes)
- perf: minor change - these resultSets will get bigger, but are still smaller than tool usage
- devops: server change is backward-compatible
- security: n/a (remembering of course that dashboard users are allowed to see secret roles)
- ux/docs: design approved by PM and UXD


model/dataSets/systemAccessDataSet.js
```diff
@@ -1,16 +1,40 @@
...
+	_sum(data) {
+		const sumMap = chain(data)
+			.groupBy('D')
+			.mapValues(dateRows => sumBy(dateRows, 'count'))
+			.value();
+
+		return Object.entries(sumMap);
+	}
+
+	_addMissingDays(data, dateRange) {
```
bemailloux, line 33: Hmm. Should these three functions be part of the Dataset base class?

bemailloux (approved):

bemailloux:
oops I approved before testing

test/server.js
```diff
@@ -32,51 +32,26 @@ const orgUnitsData = () => {
...
+	{ D: '2021-11-24', r: 580, count: 131 },
+	{ D: '2021-11-25', r: 580, count: 50 },
+	{ D: '2021-11-25', r: 580, count: 10 },
+	{ D: '2021-11-25', r: 580, count: 50 },
+	{ D: '2021-11-26', r: 580, count: 90 },
+	{ D: '2021-11-27', r: 580, count: 95 },
+	{ D: '2021-11-28', r: 580, count: 131 },
+	{ D: '2021-11-29', r: 580, count: 110 },
+	{ D: '2021-11-30', r: 580, count: 90 },
+	{ D: '2021-12-01', r: 580, count: 95 }
```
bemailloux, line 54: Suggestion: for this and E/W leave out a day to ensure "0" works

model/dataSets/systemAccessDataSet.js
johngwilkinson: The filtering will be different (different field name). The other two could just be utilities I guess. They're small enough that I thought the "refactor if it happens a third time" rule would be good.

johngwilkinson:
> oops I approved before testing

When you've tested, can you check off the box in the description? (This is both a request and a question 😃)

johngwilkinson:
> > > oops I approved before testing
> > 
> > 
> > When you've tested, can you check off the box in the description? (This is both a request and a question 😃)
> 
> I tested and found an issue. Is there a process for that?

I guess comment here with the issue? Is that image above the issue? I think I need some words too to understand the problem :)

bemailloux:
> I guess comment here with the issue? Is that image above the issue? I think I need some words too to understand the problem :)

If there is a date with no entries in the system access data, y-axis bounds are wrong

bemailloux:
> review

Should I also write down what / how I tested?

bemailloux (approved):
LGTM!

Testing notes: E2E on draco stack
* Back-end
  * Verified that data returned by API was the same as result of running query manually on the DB
  * Data for D2L support, D2L monitor, and impersonation accesses not included in API result
  * (Had to modify data slightly to get withdrawals data to appear. See back-end PR for details)
* Front-end
  * Displayed counts are correct for system access and E/W

NicholasHallman (approved):
Pulled, clicking chart items in the role tool chart filters data great across the other charts.

johngwilkinson:
Did a smoke-test on demo and e2e after the merge

consts.js
```diff
@@ -3,6 +3,9 @@ export const CHART_WIDTH_DOUBLE = 1148;
 export const CHART_HEIGHT = 260;
 export const CHART_TITLE_HEIGHT = 25;
 
+export const INACTIVE_OPACITY = 0.2;
+export const ACTIVE_OPACITY = 1;
```
NicholasHallman, line 7: Nice!

bemailloux (commented):
Code LGTM - doing another round of testing since this has changed

bemailloux (approved):
LGTM! Testing notes:
* Clicked on 1/multiple/all roles in Course access
  * System access, TUT, TUC charts were filtered
  * Course access & TUR used opacity to show selected and "hide" unselected
  * URL state changed; back/forward/reload work
  * When all roles are selected, it clears the filter
* Did a subset of the same tests on TUR
* Export
  * If there is no data in a chart, the header isn't included in the file either (I think this is a bug?)
  * Only data after filtering is included
  * For export, Course Access and TUR actually filter the data instead of including all of it

johngwilkinson:
> If there is no data in a chart, the header isn't included in the file either (I think this is a bug?)
I've noticed this too and agree it's a bug

> When all roles are selected, it clears the filter
Which I'm very proud of, but FYI this behaviour will be removed in my next PR based on designer/PM discussion yesterday

NicholasHallman (approved):

PR #144, opened 2022-04-04
US137399 Tool filter - component

bemailloux:
This PR just adds the filter component to the page and populates it with the list of tools, sorted in alphabetical order. It doesn't apply the filter yet, or add it to the URL.

Quality notes
* Testing (demo & draco-stack)
  * While tool usage trend dataset is loading, filter displays "No available filters"
  * After loading is done, tools are displayed in alphabetical order
* Security, docs, devops: N/A
* Perf
  * Made `tools` computed on the data class, so it doesn't have to recalculate as long as the data remains the same
  * It'll be interesting to figure out how this will interact with tools appearing and disappearing based on different OUs being selected. I'm not really worrying about that in this PR though.
* UX:
  * Accessibility: 
    * keyboard navigable. In fact, it comes with instructions on first use
    * MS narrator and NVDA read filter options out; dropdown button says "No filters applied" or "N filter(s) applied" depending on the number selected
  * There is a slight design mismatch with the existing OU filter. By default the Tool filter text says "Tool (n)" to indicate that there are n filters selected, and it doesn't look like this is configurable. This is different from existing dashboard filters which say "Tool: n selected"

image


model/data.js
```diff
@@ -100,6 +101,18 @@ export class Data {
 		}
 	}
 
+	get tools() {
+		const toolResultSet = this.getResultSet(resultSets.TOOL_USAGE_TREND);
```
johngwilkinson, line 105: Ah, interesting - I had to do something similar to get the available roles. I used a legend dataset instead, which simplified the code here for me - not sure if that would help you or not (no change needed)

johngwilkinson (approved):
> keyboard navigable. In fact, it comes with instructions on first use

Neat!

Design question: are we sure we don't want to show, in this filter, all available tools all the time, i.e. a fixed list that could be hard-coded (and localized) in the client?

johngwilkinson:
Sorry, forgot to test!

model/data.js
bemailloux: Wait so you used a legend dataset to populate the top-level role filter? Or was dataset used for the role interactions applied filter instead?

(I'm treating the top-level filters as separate from the filtering that happens when you click on the graphs themselves. It's weird IMO that there are multiple different filters that affect the same dimensions, but I guess that's what product wanted 🤷 )

bemailloux:
> Design question: are we sure we don't want to show, in this filter, all available tools all the time, i.e. a fixed list that could be hard-coded (and localized) in the client?

That would be simpler to implement for sure (in terms of e.g. URL state management), but could get awkward if we wanted to support additional tools. What if we made the tool list static on the back-end? Then there would only be a single source of truth for tools, and URL state management would be a bit easier. (But we'd have to solve the URL state mgmt problem for roles anyway, so :shrug:)

johngwilkinson:
Testing:
- responsive: I like how it pops up a panel on smaller screens
- rtl looks good (except narrow-screen panel is still on the right - not a big deal)
- screenreader/keyboard nav works (initially I tested with an earlier version which had a bug - but `npm update` fixed it)
- select/deselect works fine; confirmed that tool ids appear in the underlying component

johngwilkinson (dismissed):

bemailloux:
> responsive: I like how it pops up a panel on smaller screens

Gaudi did a great job building a true drop-in component here. Like, I didn't have to do anything at all to get that functionality.

Minor concern: if we put "role" in a separate filter component it might create 2 filter panels :)

model/data.js
johngwilkinson: I'm just using the list of available role ids to clear the role filter when all are selected. This is for clicking on roles in the chart, but I suppose the same dataset would work well for the top-level filter too. It just has names and ids in an array - not sure if the tools one is the same or not.

I think the top-level filters are just another interface to the same underlying filtering, right? So, if I select a role by clicking, it will also be selected in the filter, and visa-versa. Seems like it would be very confusing otherwise.

johngwilkinson:
> > Design question: are we sure we don't want to show, in this filter, all available tools all the time, i.e. a fixed list that could be hard-coded (and localized) in the client?
> 
> That would be simpler to implement for sure (in terms of e.g. URL state management), but could get awkward if we wanted to support additional tools. What if we made the tool list static on the back-end? Then there would only be a single source of truth for tools, and URL state management would be a bit easier. (But we'd have to solve the URL state mgmt problem for roles anyway, so 🤷)

In fact, the server seems to be providing a complete list of tools already: I see more tools in the tool filter on draco than I see in the tool legend. So maybe I should change my question to "Is it correct that all available tools are shown in the filter even if they are not in the data?" I think the answer is probably "yes", but that's a PM choice.

But backing up, I didn't realize the list of available tools would complicate the url state - wouldn't that just have a list of the currently-selected tool ids?

model/data.js
bemailloux: Binu into the issue that the filter needs to be aware of where it was applied from. There's a requirement that e.g. clicking on "Classlist" in Tool Usage Trend causes all the other tools in TUT to be made opaque. All the tool data still needs to be present in TUT for that to happen. But for the other cards, non-classlist data would be completely filtered out. 

I'm having a bit of a hard time imagining all the possible scenarios / test cases we'd need to consider in that case. Not that having separate filters is much simpler in terms of UX, but might be easier to reason about possible scenarios.

bemailloux:
> But backing up, I didn't realize the list of available tools would complicate the url state - wouldn't that just have a list of the currently-selected tool ids?

I'm thinking about the case where a tool is selected in URL state, but doesn't show up in the data (and tools list is not hardcoded). This situation will happen with roles too. How would we apply a filter that we can't properly show in the filter UI?
- Maybe we just throw the tool/role out? Seems like bad UX
- Maybe we add the tool/role to the server call and somehow return it in the tools or roles map, even if there isn't any data for it?

model/data.js
bemailloux: After running through scenarios for a bit - it seems like the biggest issue is the requirement (not sure if it really was a requirement or more of an off-hand comment) that selecting a tool in TUT _filters out_ that tool from TUC data. It would be much simpler if selecting a tool from TUT or TUC or the tool filter applied consistent behaviour to TUT and TUC.

So my thought is this: if a tool is selected from anywhere (TUT, TUC, tool filter):
- TUT and TUC show selected tools in full colour and non-selected in opaque
- TUR actually filters out that data

and the opposite for role filter: if a role is selected from anywhere (Course access, TUR, role filter):
- TUT and TUC filter out that data
- TUR and course access show selected in full colour and non-selected in opaque

(not sure if I'm explaining this very well?)

Vitalii-Misechko (approved):
Testing on localhost
* Responsive. Looks good except RTL version. I believe this is a component issue
* Accessibility. 
  * Opening a drop-down with space or enter key selects the first item. I guess this is okay (?)
  * Navigation with up/down doesn't work / work unstable (FF + narator). Do we support that pair?

image

bemailloux:
> Looks good except RTL version

I mean, looks like we have bigger problems in the RTL x responsive case :) maybe it's worth tackling those in its own story.

image

> Navigation with up/down keys doesn't work / work unstable (FF + narator). Do we support that pair?

Can confirm MS narrator doesn't really work at all with firefox. After tabbing around the page for a minute the whole page just freezes. I think this is narrator's fault?

johngwilkinson:
> I see that new tests comments appear in the latest PRs. It seems that I missed this new rule.

@Vitalii-Misechko 
I thought Anthony forwarded you the email... In any case, in retro we decided that for dashboards and bdp-query, code reviewers will also test the changes themselves. Trying to have fewer escaped defects. This is also why I've been pulling current PR code for bdp-query into a qa branch and deploying on draco stack.

johngwilkinson (approved):

Vitalii-Misechko (approved):
LGTM!

PR #156, opened 2022-04-07
fix missing csv headers when no data

johngwilkinson:
Quality Notes:
- tested demo and e2e
  - on draco, set date range to one day so there was no data at all and checked each csv by opening in Excel
  - set to longer w. role filter and checked a couple of csvs that have data and one that doesn't
- solution assumes our header names don't need escaping, which seems like a good bet


bemailloux:
> solution assumes our header names don't need escaping, which seems like a good bet

I think this is also true of other languages 👍

bemailloux (commented):
Something seems not quite right in the Demo page: if you select role=TA, TUT has data but TUR and TUC don't, and exporting doesn't include any data

johngwilkinson:
> > solution assumes our header names don't need escaping, which seems like a good bet
> 
> I think this is also true of other languages 👍

Oh, good thinking. Welsh has apostrophes here an there, for instance "Tynnu'n ôl" for "Witdrawals". I plugged that in to my en.js and tried it just to be sure: worked fine.

bemailloux (approved):
LGTM - tested on Draco stack
* Tested when all datasets either had or did not have data; headers appeared in the no data case and headers + data appeared in the data case

bemailloux:
> if you select role=TA, TUT has data but TUR and TUC don't, and exporting doesn't include any data

Image:
role-select

Vitalii-Misechko (approved):
Tests:
e2e on localhost. Export CSV exports data with headers. Managed to export only headers for System Access and Tool Usage. The number of exported rows for Course Access is correct.

e2e on draco: Selected `End User 1` for Physics course on Course Access. Tool Usage CSV contains Physics and **draco (6606)**. Why do we see draco in export?


e2e on draco: specified wrong role id in URL. all CSV except 'Enrollments and Withdrawals' contain only headers

johngwilkinson:
> e2e on draco: Selected `End User 1` for Physics course on Course Access. Tool Usage CSV contains Physics and **draco (6606)**. Why do we see draco in export?

The server data include rows for the current org unit (root, i.e. 'draco', by default) and its children. The tool usage dataset writes all rows the match the role filter. I _think_ that's best, but I could be wrong.

Vitalii-Misechko:
@johngwilkinson 

> > e2e on draco: Selected `End User 1` for Physics course on Course Access. Tool Usage CSV contains Physics and **draco (6606)**. Why do we see draco in export?
> 
> The server data include rows for the current org unit (root, i.e. 'draco', by default) and its children. The tool usage dataset writes all rows the match the role filter. I _think_ that's best, but I could be wrong.

I don't see anything wrong with such an approach. I just was confused for a moment when saw two lines in CSV while seeing only one line on the chart.

PR #157, opened 2022-04-07
fix chart shows data when all is filtered

johngwilkinson:
See https://github.com/Brightspace/insights-adoption-dashboard/pull/156#issuecomment-1091904240 for a recording of the problem.

Although the data was being passed to Highcharts correctly, the empty case series was not quite correct and this resulted in Highcharts leaving behind one series of data when the role filter should have resulted in no visible results.

Quality Notes
- functional: e2e and demo


bemailloux (dismissed):
LGTM! tested various combinations of role selections on demo page, as well as reloading

Vitalii-Misechko (dismissed):
Thank you for the fix!
Tests: Tool Usage Trend by Tool doesn't show data after applying filters

Vitalii-Misechko (approved):

bemailloux (approved):

PR #165, opened 2022-04-08
set date filter on demo page to a known value for consistent diffs

johngwilkinson:
Puppeteer doesn't directly support date manipulation, and importing sinon to use fake timers would be a bit hacky.

So I went with using the demo flag. Not sure if this is the very best way, but it works well enough.

Quality:
- checked demo page, played with date filter
- smoke-test on draco stack
- generated some visual diffs locally with the date set back in time; confirmed the older date appears in the images
- devops: using today's date, so if this goes in today, any diffs already generated today will still be good


Vitalii-Misechko (approved):
tests:
* draco env: End Date shows the current date
* local env: End Date shows the date specified for the demo environment

NicholasHallman (approved):
LGTM
- demo uses hardcoded
- e2e uses today

PR #170, opened 2022-04-11
Us136840 Series colors

bbabu01:
Map colors to each series inside corresponding legend dataset
In another PR: Patterns

Quality notes:

- Testing:

  - Retain previous mapping for same series across multiple loads with and without additional series in new load
  - Mapping when there are more series than number of colors
  - Mapping when there are more colors than number of series
  - Interactions test - Series should retain color before and after interactions filter the dataset. eg: Click on a series on TUT and verify TUR and TUC 
- Security, devops, docs, perf: N/A
- Unit tests - Pending
- UX:
  - [ ] Demo


components/cards/bar-chart-with-drill-card.js
```diff
@@ -241,16 +243,13 @@ class BarChartWithDrillCard extends SkeletonMixin(Localizer(MobxLitElement)) {
...
-		return data.map(({ id, name, data, isDeselected }, idx) => ({
+		return data.map(({ id, name, data, isDeselected }) => ({
 			id: id,
 			name: name || id,
 			data,
 			// the property is "isDeselected": series appear selected by default
 			opacity: isDeselected ? INACTIVE_OPACITY : ACTIVE_OPACITY,
-			color: colors[idx] // this will match the rest of the legend as long as we order by name
-			// TODO: figure out a better colouring system
+			color: this.legendDataSet?.data?.find(item => item.name === name)?.color,
```
johngwilkinson, line 252: Can you match on `id` instead of name? It's possible for two org units to have the same name

Just want to note in passing that this is an O(n^n) algorithm (search the list of org units for each org unit), which could be a big problem for large N. Since we limit to 200, I'm not concerned - although since there's already a Map of colours internally in the legend dataset, it might make sense to pass that along and use it here.

model/dataSets/roleLegendDataSet.js
```diff
@@ -1,21 +1,36 @@
...
+	}
+
 	process([toolUsageRoles, courseAccessRoles]) {
 		const dedupedRoles = new Map([...courseAccessRoles.series, ...toolUsageRoles]
 			.sort(compareNameAndId)
 			.map(({ id, name }) => [Number(id), name])
 		);
-		return [...dedupedRoles].map(([id, name]) => ({ id, name }));
+
+		const existingRoles = [...this.roleColorMap.keys()];
```
johngwilkinson, line 23: I don't think you need this for roles or tools, because the tool usage ResultSet contains all possible roles and tools. The only reason for the deduping from two datasets here is that the legend should show only roles that are actually in the data. But you can stably (for a given tenant) assign colours using the role map in the ResultSet, I think.

bbabu01: Got it. Wonder if we should account for a  scenario in which dashboard only gets courseaccess data.

johngwilkinson: If it's due to an error, I don't think it matters. But it's worth double-checking that the role and tool lists come back even if there are no rows in the tool usage data

PR #175, opened 2022-04-12
US135379: [Adoption] [Date Range Filter] Filter UI

Sean-Kurian:
[US135379](https://rally1.rallydev.com/#/609774592091d/teamboard?detail=%2Fuserstory%2F622721217831): [Adoption] [Date Range Filter] Filter UI
[US137614](https://rally1.rallydev.com/#/609774592091d/teamboard?detail=/userstory/630757044415): [Adoption] Move Date Filters

* Functional Testing 
  * [x] Browsers (Chrome, Firefox, Edge)
  * [x] Accessibility
  * [x] Responsive
  * [x] Test cases
    * [x] Date range syncs with URL for bookmarking purposes
    * [x] Can no longer bypass enforced date range using 'Today' button 
* Security, devops, perf: N/A
* UX and docs

* [x] Demo


pages/adoption-dashboard-view.js
```diff
@@ -50,8 +50,7 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 		super();
 		this.data = null;
 		this.ouFilterDataManager = null;
-		const lowerDateBound = currentTime();
-		lowerDateBound.setMonth(lowerDateBound.getMonth() - 6);
+		const lowerDateBound = new Date(Date.parse(new Date()) - 6 * 30 * 24 * 60 * 60 * 1000);
 		this._minValue = formatDate(lowerDateBound, { format: 'yyyy-MM-dd' });
```
johngwilkinson, line 54: merge issue here: the fix for the vis diff dates was to use `currentTime()`, not `new Date()` or `Date.now()`

pages/adoption-dashboard-view.js
```diff
@@ -163,13 +162,9 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 			}
 			.d2l-insights-filter-layouts {
 				display: flex;
-				flex-wrap: wrap;
 				justify-content: space-between;
 				margin-top: -5px;
-				width: calc(100vw - 80px);
-			}
-			.d2l-dimension-filter-group {
```
johngwilkinson, line 171: maybe a merge issue here? Looks like the style for `.d2l-dimension-filter-group` got deleted

pages/adoption-dashboard-view.js
```diff
@@ -204,6 +199,8 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 		const toolUsageComparisonDataSet = this.data.getDataSet(dataSets.TOOL_USAGE_COMPARISON);
 		const toolsLegendDataSet = this.data.getDataSet(dataSets.TOOL_USAGE_TREND_BY_TOOL_LEGEND);
 		const rolesLegendDataSet = this.data.getDataSet(dataSets.TOOL_USAGE_TREND_BY_ROLE_LEGEND);
+		const today = formatDate(new Date(), { format: 'yyyy-MM-dd' });
```
johngwilkinson, line 202: merge issue: these lines shouldn't be here (they are now modified and below in `renderFilters`)

pages/adoption-dashboard-view.js
```diff
@@ -217,7 +214,22 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 		return html`
 			${this.renderHeader()}
 			<div class="d2l-insights-sections-layout">
-				${this.renderFilters()}
```
johngwilkinson, line 220: merge issue: this code got moved down to `renderFilters`

model/data.js
```diff
@@ -21,15 +21,16 @@ export class Data {
 		this._dataApiClient = dataApiClient;
 		this._allError = false; // @observable
 		this._dataSetFactories = dataSetFactories;
-		this._dateRange = undefined;
+		this._dateRange = JSON.stringify({ startDay: '2022-02-25', endDay: '2022-03-10' });
```
johngwilkinson, line 24: The default should be an empty string. Otherwise when the dashboard first loads, this param gets added to the url and we get a little bug where hitting "back" doesn't go back to the Insights Portal until you hit it twice.

Also, see the `decorate` block at the bottom of this file: `dateRange` should be added there as `computed`, and `_dateRange` as `observable`.

model/data.js
```diff
@@ -21,15 +21,16 @@ export class Data {
...
 		this._orgUnitDrillPath = '';
 		syncWithUrl(this, 'orgUnitDrillParam', orgUnitDrillPathParam);
+		syncWithUrl(this, 'dateRange', this._dateRange);
```
johngwilkinson, line 27: The third parameter here will show up as the key in the url, as in `key=value`. It needs to be a short constant string, like `orgUnitDrillPathParam` is. Notice what shows up in the url with this code, and also that changing the date then hitting "back" results in an error.

model/data.js
```diff
@@ -21,15 +21,16 @@ export class Data {
...
 		this.roleFilter = new RoleFilter(this);
 	}
 
 	load(alert) {
 		this._allError = false;
-		this._dataApiClient.startDataRequest({ parentId: this.parentId, dateRange: this._dateRange });
+		this._dataApiClient.startDataRequest({ parentId: this.parentId, dateRange: JSON.parse(this._dateRange) });
```
johngwilkinson, line 33: you should use `this.dateRange` instead of repeating `JSON.parse...` here

model/data.js
```diff
@@ -58,14 +59,14 @@ export class Data {
 		return this._resultSets.some(r => r.isQueryError);
 	}
 
-	getDateRange() {
-		return this._dateRange;
+	get dateRange() {
+		return JSON.parse(this._dateRange);
```
johngwilkinson, line 63: update to handle the default case (null/undefined or empty string, see above comment). You could return `null` and then use the start/end dates calculated in the view; or you could calculate the default start/end dates here (including the `currentDate` stuff)

model/data.js
```diff
@@ -58,14 +59,14 @@ export class Data {
...
 	}
 
-	setDateRange(alert, dateRange) {
-		if (typeof this._dateRange === 'undefined' || this._dateRange.startDay !== dateRange.startDay || this.dateRange.endDay !== dateRange.endDay) {
-			this._dateRange = { startDay: dateRange.startDay, endDay: dateRange.endDay };
-			this.load(alert);
+	set dateRange(dateRange) {
+		if (this._dateRange === '' || this._dateRange.startDay !== dateRange.startDay || this._dateRange.endDay !== dateRange.endDay) {
+			this._dateRange = JSON.stringify(dateRange);
```
johngwilkinson, line 68: In general, I would use another get/set pair for the stringified value (`dateRangeParam`) and give that one to `syncWithUrl`, then store the dateRange object in `this._dateRange`. The current way incurs some extra calculation because it has to parse the string in order to use it, although it's not really enough to worry about.

Another suggestion: seeing unlencoded JSON in the url is a bit odd. Since really this is two parameters (startDay and endDay), why not provide a getter and setter for each? So a `get startDay() { return this._dateRange.startDay; }` and similar for `endDay`. Then in the constructor you have
```
syncWithUrl(this, 'startDay', startDayUrlParamName);
syncWithUrl(this, 'endDay', endDayUrlParamName);
```
And add both as `calculated` at the bottom of this file
```

model/data.js
johngwilkinson, line 27: The second parameter here has to be a string, not an object. That would be `_dateRange` currently.

Sean-Kurian:
Making a new branch for required changes and to fix merge issues, please don't review this PR any further

PR #183, opened 2022-04-19
US135379: [Adoption] [Date Range Filter] Filter UI Fixes

Sean-Kurian:
[US135379](https://rally1.rallydev.com/#/609774592091d/teamboard?detail=%2Fuserstory%2F622721217831): [Adoption] [Date Range Filter] Filter UI
[US137614](https://rally1.rallydev.com/#/609774592091d/teamboard?detail=/userstory/630757044415): [Adoption] Move Date Filters

* Functional Testing 
  * [x] Browsers (Chrome, Edge)
  * [x] Accessibility
  * [x] Responsive
  * [x] Test cases
    * [x] User will have the ability to open filter and select dates
    * [x] Date defaults to 14 day range, at most can be 6 months ago to present
    * [x] Changing URL to >6 month date range loads default data
* Security, devops, perf: N/A
* UX and docs

* [x] Demo


pages/adoption-dashboard-view.js
```diff
@@ -50,10 +54,13 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 		this.data = null;
 		this.ouFilterDataManager = null;
-		const lowerDateBound = currentTime();
-		lowerDateBound.setMonth(lowerDateBound.getMonth() - 6);
+		const lowerDateBound = new Date(Date.parse(currentTime()) - 6 * 30 * 24 * 60 * 60 * 1000);
 		this._minValue = formatDate(lowerDateBound, { format: 'yyyy-MM-dd' });
 		this._maxValue = formatDate(currentTime(), { format: 'yyyy-MM-dd' });
+		this._startDay = '';
+		this._endDay = '';
+		syncWithUrl(this, 'startDay', startDayPathParam);
```
johngwilkinson, line 62: Can these be moved into the `Data` class please, like the drill path?

insights-adoption-dashboard.js
```diff
@@ -130,7 +130,17 @@ class AdoptionDashboard extends Localizer(MobxLitElement) {
 				dataSetFactories,
 				dataApiClient: new DataApiClient(this.metronEndpoint)
 			});
-			this._data.load(this, undefined);
+
+			const urlsp = new URLSearchParams(location.search);
```
johngwilkinson, line 134: No need for this: when the synced params are in the Data class, the `load()` method can just read them directly. Same as the drill path.

pages/adoption-dashboard-view.js
```diff
@@ -168,7 +175,7 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 				flex-wrap: wrap;
 				justify-content: space-between;
 				margin-top: -5px;
-				max-width: 1200px;
+				width: calc(100vw - 165px);
```
johngwilkinson, line 178: Is this still needed? I noticed Nick made some changes to layout that might have solved the same problem, and the diffs make it look like this change has misaligned the date filter - I'm not sure though.

pages/adoption-dashboard-view.js
Sean-Kurian: If this was moved to the `Data` class, would I still be able to set the initial values of the component itself to the URL values? As an instance of `Data` isn't initialized until after the constructor call.

johngwilkinson: The component values are set as properties on the component right? Like (pseudocode) `<date-filter start="${start}" end="${end}"></date-filter>`. So you just need to get those values from `data`. Just like for the course access card, for example, we have `.drillPath="${this.data.orgUnitDrillPath}"`

In general in Lit, you don't need to set anything of interest in the constructor: it's the `render()` method that matters.

pages/adoption-dashboard-view.js
```diff
@@ -169,14 +168,17 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
...
 			.d2l-dimension-filter-group {
 				margin-bottom: 50px;
 			}
 			d2l-input-date-range {
 				position: relative;
 				top: -28px;
-				width: 332px;
+				width: 400px;
+				display: flex;
+				justify-content: end;
```
NicholasHallman, line 181: I've added these changes because occasionally the browser would stack the date filters if we gave the perfect width. The date components don't seem to size themselves appropriately immediately so the flex box and wiggle room helps keep them lined up

pages/adoption-dashboard-view.js
```diff
@@ -441,25 +446,52 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+			});
+			el[1].requestValidate(true);
+
+			this._minValue = formatDate(lowerDateBound, { format: 'yyyy-MM-dd' });
+			this._maxValue = upperDateBound > new Date() ? formatDate(new Date(), { format: 'yyyy-MM-dd' }) : formatDate(upperDateBound, { format: 'yyyy-MM-dd' });
+
+			const startDay = new Date(startVal);
+			const endDay = todayDateString.toDateString() === new Date().toDateString() ? null : new Date(endVal);
+
+			this.data.setDateRange(this, { startDay: startDay, endDay: endDay });
```
johngwilkinson, line 482: You're tracking the date range and the start and end days as two separate things. There should only be one underlying variable in Data, and the other should be calculated from it. I'm thinking maybe the best thing would be to get this merged though, then I can refactor a bit and give you a tour of the results.

pages/adoption-dashboard-view.js
```diff
@@ -441,25 +446,52 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		const endVal = target.endValue;
+		const startVal = target.startValue;
+
+		const lowerDateBound = new Date(Date.parse(endVal) - 6 * 30 * 24 * 60 * 60 * 1000);
+		const upperDateBound = new Date(Date.parse(startVal) + 6 * 30 * 24 * 60 * 60 * 1000);
+		if (Date.parse(endVal) <= Date.parse(upperDateBound) && Date.parse(startVal) >= Date.parse(lowerDateBound) && Date.parse(startVal) <= Date.parse(endVal)) {
+			const todayDateString = new Date(endVal);
+
+			todayDateString.setDate(todayDateString.getDate() + 1);
+			lowerDateBound.setDate(lowerDateBound.getDate() + 1); //accounting for JS date casting setting one day back due to GMT-4
```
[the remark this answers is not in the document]
Sean-Kurian: I actually think this is always correct? As the what's happening here is that endVal comes from the browsers time zone, but is just a raw date - like '2022-04-21', which is then cast into a date using new Date() - but this always returns a day before when cast like this because that date is converted to GMT -4. The initial time comes from the browser itself so it will always be the correct day, right?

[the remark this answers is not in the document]
johngwilkinson: Why GMT-4? Only because that's your browser's time zone. Here are some examples; notice that the actual time is midnight UTC, but that gets adjusted so that it could be a day before _or not_ depending on the time zone. Notice also that using `new Date(year, month -1, day) actually gives midnight in the browser's time zone, if that's what you need:
image

(Notice you don't need `new Date(Date.parse(d))`, btw: `new Date(d)` will do)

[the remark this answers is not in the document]
Sean-Kurian: I added the timezone offset to each date to resolve this issue - now theyre all unified and being compared in the correct time zones.

johngwilkinson (approved):
Okay, let's get this merged and then I'll do a bit of refactoring.

Testing:
- changing the date loads as expected
- refreshing the page works, with both default and specified end date
  - interesting side-effect: there's no way to make a bookmark for a fixed date range that includes today's date; not sure if that's something we should change or not
- browser back/forward buttons update the date but don't reload the data - should be fixed by the changes I'm planning

NicholasHallman (approved):
Testing
 - refreshed page with only start moved 
 - refreshed page with on end moved
 - refreshed page with both
 - set start to after end and refreshed, didn't break
 - date changes loads proper data

PR #191, opened 2022-04-22
US135379: bug fixes and date refactor

johngwilkinson:


Vitalii-Misechko (approved):
Tests:

* the first navigation to the page shows a valid Start Date that is two weeks ago
* setting an invalid EndDate from the dropdown shows a warning icon but it also reloads the dashboard
  *  should we prevent fetching data for this case? or use `intercept` to cancel MobX reaction?
* dates are correctly persisted/restored to/from URL
* incorrect dates that are restored from URL show warning icons
  * setting focus to a date control causes dashboard reload.
  * should we just clear these dates instead?

Vitalii-Misechko:
> > * incorrect dates that are restored from URL show warning icons
> 
> Can you give an example? I tried a date like 2022-04-0x (not a valid date string), and didn't see the warning, but the date input appeared empty until I clicked on it. I'd like to fix this too.

I typed 2022-**20**-27 in URL for the start date parameter and then clicked inside the input of Start Date. After that dashboard reloaded

image


But your latest change fixed this.

Tests:

* the first navigation to the page shows a valid Start Date that is two weeks ago
* setting an invalid EndDate from the dropdown shows a warning icon but it also reloads the dashboard
  *  fixed
* dates are correctly persisted/restored to/from URL
* incorrect dates that are restored from URL get corrected

Vitalii-Misechko (approved):
LGTM!

PR #193, opened 2022-04-22
Colors pattern fill

bbabu01:
Use a chart as legend
Pattern fill role colors and legends


components/cards/bar-chart-with-drill-card.js
```diff
@@ -220,6 +222,17 @@ class BarChartWithDrillCard extends SkeletonMixin(Localizer(MobxLitElement)) {
 								that.toggleSeries(this.series?.userOptions?.id);
 							}
 						}
+					},
+					color: {
+						pattern: {
```
johngwilkinson, line 227: This `pattern` object is identical in all three places, right? So making a constant for it would be good.

NicholasHallman:
Can we make the colored bars in the pattern thicker and the white space thinner? It looks dithered atm which will make like colors seem more similar

bbabu01:
> Visual diffs need updating

Thanks. Looking for a way to fix the rtl issues

components/cards/legend-card.js
```diff
@@ -0,0 +1,169 @@
...
+				} : this.colorMap?.data?.get(id)
+			};
+		}
+		);
+	}
+}
+decorate(LegendCard, {
+	_series: computed
+});
+customElements.define('d2l-insights-chart-legend-new', LegendCard);
```
bemailloux, line 169: Are we still using the old legend? If not, just delete it and call this one `d2l-insights-chart-legend`

bemailloux (commented):
Some notes about the visual diffs:
- Legend should use the "Lato" font if possible (to match everything else)
- In the "Course access data error" visual diff, the roles legend is missing even though TUR is showing

bbabu01:
> Some notes about the visual diffs:
> 
> * Legend should use the "Lato" font if possible (to match everything else)
> * In the "Course access data error" visual diff, the roles legend is missing even though TUR is showing

Checking this. Looks like this might have been broken for a while

components/cards/legend-card.js
```diff
@@ -0,0 +1,177 @@
...
+			></d2l-labs-chart>`;
+	}
+
+	get _error() {
+		return this.dataSet.isQueryError;
+	}
+
+	get _chartOptions() {
+		let direction;
+		(document.dir !== undefined) ? direction = document.dir : direction = document.getElementsByTagName('html')[0].getAttribute('dir');
```
bemailloux, line 82: A more idiomatic way to write this would be 
```
direction = document.dir !== undefined ? document.dir : document.getElementsByTagName('html')[0].getAttribute('dir');
```

That way it's obvious that there's a variable assignment happening here.

(Of course you could then further simplify to 
```
direction = document.dir || document.getElementsByTagName("html")[0].getAttribute('dir');
```
)

bemailloux (commented):
Code LGTM - testing now

bemailloux:
This is interesting: a scroll bar appears on the legends when they're loading. That's probably fine, just wondering if you noticed that.

image

bemailloux (approved):
LGTM! Testing notes
* Loaded dashboard multiple times with different dates and drills, roles colour/pattern remained constant across loads
* Pattern scales when page is zoomed in/out
* Tested on Chrome / FF / Edge
* Pattern looks good in PDF page printout (even if layout is kind of messed up)
* Legend colour circles could probably be a bit larger IMO

bbabu01:
> This is interesting: a scroll bar appears on the legends when they're loading. That's probably fine, just wondering if you noticed that.
> 
> image

Yes, noticed that but couldn't find a way to prevent it since the container height is already more than charts

bemailloux (dismissed):

bbabu01:
> LGTM! Testing notes
> 
> * Loaded dashboard multiple times with different dates and drills, roles colour/pattern remained constant across loads
> * Pattern scales when page is zoomed in/out
> * Tested on Chrome / FF / Edge
> * Pattern looks good in PDF page printout (even if layout is kind of messed up)
> * Legend colour circles could probably be a bit larger IMO

Thanks. Increased the size. Also noticed a tiny fade on hover and fixed it too.

bemailloux (approved):

PR #198, opened 2022-04-26
US135379: if end day is today, send a null instead (fix regression)

johngwilkinson:
My earlier changes started sending a date when the end date is today's date instead of `null`. The `null` is needed for cases where there may be server data for "tomorrow", e.g. the server is in Australia but my browser is in Ontario.

testing (e2e)
- load with no query params: expected date range appears, and end date is sent to server as null
- change start date: expected dates appear in url (start and end), and end date is sent to server as null
  - reload: loads same
- change end date: appears in url, both start and end dates are sent to the server
  - reload: loads same


Vitalii-Misechko (commented):
Tests:

* initial load. The end date is null
* change the start date. The end date is null. Dates are persisted in URL
* change the end date. The end date is not null. Dates are persisted in URL
* copy a URL in a new window when the end date is today. The end date is null. Dates are persisted in URL
* ~copy a URL in a new window when the end date is yesterday. The end date is null. Dates are persisted in URL. **Why the end date is null in this case?**~
* copy a URL in a new window when the end date is yesterday. The end date is yesterday. Dates are persisted in URL.

Vitalii-Misechko (dismissed):

bemailloux (approved):
LGTM! Testing notes
- Set timezone to UTC-4, sent "null" for today (2022-04-27), sent not null for yesterday
- Set timezone to UTC+14, sent "null" for today (2022-04-28), sent not null for yesterday and the day before
- Set timezone to UTC-12, sent "null" for today (2022-04-27), sent not null for yesterday
- Set end date to today's date using URL params, today was still sent as null
- I'm not able to test the situation where browser's today is the day before server's today (can't change the time on my PC). Maybe we could test this by changing draco-stack's timezone, but that has side effects.

Vitalii-Misechko (approved):

bemailloux (approved):

PR #213, opened 2022-05-11
US132923 [Adoption][Org Filter] Apply Tree Filter Selection API

Vitalii-Misechko:

* Functional Testing 
  * [x] E2E Test cases
    * [x] Search skips invisible org units that match query string
    * [x] Search is pageable
    * [x] Expanding a tree node show Load more if not all nodes are loaded
    * [x] Telemetry works
      * [x] Dashboard works when telemetry is on 
* Security, devops, perf: N/A
* UX and docs: N/A

* [ ] Demo


Vitalii-Misechko:
> Code looks good. Is it worth waiting a day or two for the LMS change to reach draco stack, and do e2e testing?

That's good. I hope to make E2E tests today with local BSI. But, I can merge it right away after approving the PR (in order to s make E2E tests on draco stack tomorrow)

PR #215, opened 2022-05-11
Add Select all button

Vitalii-Misechko:

* Get Select all button visible. 
* Get child nodes selected instead of the root node (the default behavior for Select all) 

Rationale
* Selected nodes are used as filters therefore we can't use root ID because we don't load data for it. Moreover, we want to see aggregation at the child level

Feature request thread https://d2l.slack.com/archives/CMNGJ22G0/p1648468237207089

E2E Tests on Draco
* [x] Select all show correct charts with aggregation at the child level
* [x] Selecting/deselection a child org unit updates charts accordingly

### Screenshot

image


johngwilkinson (approved):
Testing (e2e draco):
- hit select all on first load (no effect except to fill in checkboxes)
  - clear some org units and see graphs update as expected
- clear; open Accounting down to the course level, hit select all
- drill, select all, clear some

PR #217, opened 2022-05-11
Clears selected org unit when a user drills up a few levels at once

Vitalii-Misechko:

* Functional Testing 
  * [x] E2E Test cases
    * [x] Selection for a give level is restored when a user drill up to that level
    * [x] Selection is cleared if a user drill up from that level
    * [x] Selection is cleared for all levels beneath the level that user drilled up to
* Security, devops, perf: N/A
* UX and docs: N/A


johngwilkinson (approved):
Tested on draco and confirmed this does what it says. Code looks good.

I wonder, though, wouldn't it be better not to clear the selections for the current and intermediate drill levels at all? That way when I drill back down, my selection is still there.

Maybe something to discuss w. PM/design during demo.

Vitalii-Misechko:
We decided on Today's standup that we want to preserve selection. Therefore, I close this PR

PR #228, opened 2022-05-17
Notify the user of truncation

bbabu01:
[US137545](https://rally1.rallydev.com/#/?detail=/userstory/630651451249&fdp=true): [Adoption] Notify the user of truncation

Draft: Display d2l-alert if more than 100 org units selected.

Tested on draco with a smaller cut ff


pages/adoption-dashboard-view.js
```diff
@@ -215,7 +218,12 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 			${this.renderHeader()}
 			<div class="d2l-insights-sections-layout">
 				${this.renderFilters()}
-
+				<d2l-alert
+					has-close-button
+					?hidden=${!this.data.ouFilter._isTruncated}
```
Vitalii-Misechko, line 223: `_` - underscore denotes a private member in JS. It seems you should get rid of it

Vitalii-Misechko:
@bbabu01 thank you for your improvements.

A few things that came up to my mind:
* Checking `this._availableOrgUnitIds` doesn't make sense, because it is just a reflection of `this.selectedOrgUnitIds` plus children of the current org unit root (drill)
  * Therefore, we should check the children number of the current root org unit `$0.dataManager.orgUnitTree.getChildIds(dataManager.orgUnitTree.rootId).length`
  * and `$0.data.ouFilter.selectedOrgUnitIds`
  * $0 above is an instance of AdoptionDashboardView
* I believe that is the best place where we can get these values `AdoptionOuFilterDataManager` or `AdoptionDashboardView`
* Your story covers the case that is described here [US139191](https://rally1.rallydev.com/#/?detail=/userstory/636950505235&fdp=true): _[Adoption][Org Filter] Load more for the current drill should also fetch corresponding analytical data_. Therefore, we don't need US139191 (my story)

Some questions
* Should we introduce a flag at the backend that signals us that a user tries to get data for more than 100 org units? (I think no, but there may be other thoughts around)
  * if yes, should we try to truncate data at the backend?
    * if yes, which data should we truncate the current org unit children or selected org units?

@johngwilkinson, here are a few thoughts on data truncation notification. Have a look, please.

pages/adoption-dashboard-view.js
```diff
@@ -503,6 +511,13 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 	_exportToCsv() {
 		this.data.export(term => this.localize(term));
 	}
+
+	_isTruncated() {
+		const availableOrgUnits = this.ouFilterDataManager.orgUnitTree.getChildIds(this.ouFilterDataManager.orgUnitTree.rootId);
+		const selectedOrgUnits = this.data.ouFilter.selectedOrgUnitIds;
+		return (availableOrgUnits.length > ORG_UNIT_TRUNCATION_LIMIT && selectedOrgUnits.size === 0) ||
```
Vitalii-Misechko, line 518: Server limits the number of children to 100. Therefore, you should use `>=` for comparison with `availableOrgUnits`

model/ouDimensionFilter.js
```diff
@@ -95,6 +96,11 @@ export class ouDimensionFilter {
 
 		this.setSelectedOrgUnitIds(newValue);
 	}
+
+	get isTruncated() {
+		return (this._availableOrgUnitIds.length > ORG_UNIT_TRUNCATION_LIMIT && this.selectedOrgUnitIds.size === 0) ||
+		this.selectedOrgUnitIds.size > ORG_UNIT_TRUNCATION_LIMIT;
+	}
```
Vitalii-Misechko, line 103: Did you just forget to delete this?

Vitalii-Misechko (commented):
1. When a selected number of nodes exceeds the limit the dashboard shows the alert and reloads data. It should not reload I think

1. You need to cover the alert with a visual-diff test I guess

pages/adoption-dashboard-view.js
Vitalii-Misechko: Sorry. It seems I'm wrong here.

If a user hits  "load more" (it appears only if there are more than 100 ous) then your code will work fine. So no need `>=` comparison

Vitalii-Misechko:
here is what I received when loaded the dashboard first time (demo page)
image

bbabu01:
> 1. When a selected number of nodes exceeds the limit the dashboard shows the alert and reloads data. It should not reload I think
> 2. You need to cover the alert with a visual-diff test I guess

Will add vis diff in a separate PR

model/ouDimensionFilter.js
```diff
@@ -79,7 +80,7 @@ export class ouDimensionFilter {
 	_shouldReload(newOrgUnitIds) {
 		// if all ous are not available without loading
 		const availableOrgUnitIds = new Set(this._availableOrgUnitIds);
-		return ![...newOrgUnitIds].every(ou => availableOrgUnitIds.has(ou));
+		return ![...newOrgUnitIds].every(ou => availableOrgUnitIds.has(ou)) &&  newOrgUnitIds.size <= ORG_UNIT_TRUNCATION_LIMIT;
```
Vitalii-Misechko, line 83: You need to add a unit test for `setSelectedOrgUnitIds` setter to cover this case in ouDimensionFilter.test.js
`nitpicking` there are two spaces after `&&` instead of one

Vitalii-Misechko:
Hi @bbabu01,
[here is PR](https://github.com/Brightspace/insights-adoption-dashboard/pull/246) that uses the backend piece of code related to Truncation. Feel free to approve/merge or add comments

test/model/ouDimensionFilter.test.js
```diff
@@ -245,5 +245,21 @@ describe('ouDimensionFilter', () => {
...
+		it('does not change orgUnitIdsToRequest if total org unit selections exceed limit', () => {
+
+			const currentDrillLevel = 6607;
+			const data = { parentId: currentDrillLevel };
+			const orgUnitResultSet = { isLoading: false, data: {
+				rootOrgUnitId: 6606,
+				requestedOrgUnitIds: [1, 2, 3]
+			} };
+
+			const sut = new ouDimensionFilter(data, orgUnitResultSet, '',  3);
```
Vitalii-Misechko, line 258: `nitpicking` could you create a constant for `3` in this test in order the test becomes a little bit more self-explanatory

pages/adoption-dashboard-view.js
```diff
@@ -191,6 +191,9 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
 			.d2l-insights-double-width-card-legend {
 				overflow: visible;
 			}
+			d2l-alert {
```
Vitalii-Misechko, line 194: I would suggest adding margin under the alert. In order to improve layout

image


```
    margin-bottom: 24px;
```

pages/adoption-dashboard-view.js
Vitalii-Misechko: I had a discussion with John [here](https://github.com/Brightspace/brightspace-data-platform/pull/106) on how this should work. Essentially, it should look like this
```JS
(isTruncatedOnServer && selectedOrgUnits.size === 0) || selectedOrgUnits.size > ORG_UNIT_TRUNCATION_LIMIT
```

Vitalii-Misechko: - `isTruncatedOnServer` indicates if the current drill has more than 100 OUs
- `isTruncatedOnServer` and `selectedOrgUnits` are all available in `ouDimentionFilter` because of that we can move implementation to `ouDimentionFilter` this way it will be easier to make unit tests

consts.js
```diff
@@ -6,6 +6,8 @@ export const CHART_TITLE_HEIGHT = 25;
 export const INACTIVE_OPACITY = 0.3;
 export const ACTIVE_OPACITY = 1;
 
+export const ORG_UNIT_TRUNCATION_LIMIT = 2;
```
Vitalii-Misechko, line 9: It seems you committed a test value

test/model/ouDimensionFilter.test.js
```diff
@@ -245,5 +245,66 @@ describe('ouDimensionFilter', () => {
...
+			expect(sut.isTruncated).to.be.false;
+		});
+
+		it('server data not truncated and number of selections more than ORG_UNIT_TRUNCATION_LIMIT', () => {
+			const currentDrillLevel = 6607;
+			const data = { parentId: currentDrillLevel };
+			const orgUnitResultSet = { isLoading: false, data: {
+				rootOrgUnitId: 6606,
+				requestedOrgUnitIds: [1, 2, 3, 4],
+				isTruncated: true
```
Vitalii-Misechko, line 301: `server data not truncated` then this line should look like `isTruncated: false`

Vitalii-Misechko:
Tests looks good!

Vitalii-Misechko:
I have added a few E2E tests to the PR description

Vitalii-Misechko (approved):
LGTM!

Vitalii-Misechko (commented):
Waiting on End-to-End tests

PR #231, opened 2022-05-19
US139107 Fix spinner under Enrollments and Withdrawals

Vitalii-Misechko:
Hides spinner under Enrollments and Withdrawals

* Functional Testing 
  * [x] Responsive
* Security, devops, perf: N/A
* UX and docs: N/A

* [x] Demo

### Demo

image


johngwilkinson:
Notes:
- I only see the skeleton for Engagements and Withdrawals - shouldn't there be a skeleton for the other two legends?
- E&W doesn't actually need a skeleton, because it's always the same legend

johngwilkinson (approved):
LGTM

PR #240, opened 2022-05-20
US138906: Display Roles Legend if either CourseAccessDataSet or ToolUsageTrendByRoleDataSet loads

saadmateen:
- Functional:
  - Locally and on Draco LMS
  - Cause courseAccessDataError, Roles legend still displays
  - Cause toolUsageTrendDataError,  Roles legend still displays
  - Cause both courseAccessDataError and toolUsageTrendDataError, Roles legend does not display in this case
- Security: n/a
- Perf: no extra server hits
- devops: n/a
- Ux/docs: n/a
- Demo: Completed

Rally: [US138906](https://rally1.rallydev.com/#/609774592091d/backlog?detail=%2Fuserstory%2F635063953803&fdp=true)

---------------------------------------
CourseAccessError

---------------------------------------

ToolsUsageTrendError

---------------------------------------

BothErrors


model/dataSet.js
```diff
@@ -56,11 +56,11 @@ export class DataSet {
 	}
 
 	get isLoading() {
-		return this._resultSets.some(r => r.isLoading);
+		return this._resultSets.every(r => r.isLoading);
```
johngwilkinson, line 59: What happens if you don't make the change on this line? It seems like `isLoading` should become false when an error arrives - which you could do on line 26 of `resultSet.js`.

The possible problem with doing it here, as in this PR, is that if one of the results arrives significantly later than the other, the user will see the legend render, then render again when more data arrives. I suspect that's not what we want.

saadmateen: I have made the change so that `isLoading` becomes false in `resultSet.js` on line 26, and changed back the isLoading getter  in `dataSet.js`.

johngwilkinson (dismissed):
This PR can be merged I think, but there is a problem that should be addressed in another PR.

I did e2e testing by having the server return an error (just added a `throw new Error('testing')` to the `postProcess` method for the workItem).
- if course access has an error, this works perfectly
- if tool usage has an error, this doesn't work

Here's why: the role filter is based on the tool usage result set (see `data.js` line 30), so when that result set fails, there are no roles in the filter, and all the course access data gets filtered out.

Possible fixes: 

1. make a new DataSet for the roles filter, using both datasets and falling back to the roles in course access if tool usage doesn't load. This isn't ideal because the roles filter will be missing some roles. But maybe it's okay for this edge case.
2. make a new result set server-side that just returns the roles - similar to the way the org unit filter has its own result set. This is a bit more work, but will likely behave best and seems more correct. It also has the advantage that the role filter will get populated as early as possible instead of waiting for tool usage data.

So I'd suggest merging this then working on solution (2). Since time is a bit tight, I could put together a PR for the server-side change if you like.

bbabu01:
> This PR can be merged I think, but there is a problem that should be addressed in another PR.
> 
> I did e2e testing by having the server return an error (just added a `throw new Error('testing')` to the `postProcess` method for the workItem).
> 
> * if course access has an error, this works perfectly
> * if tool usage has an error, this doesn't work
> 
> Here's why: the role filter is based on the tool usage result set (see `data.js` line 30), so when that result set fails, there are no roles in the filter, and all the course access data gets filtered out.
> 
> Possible fixes:
> 
> 1. make a new DataSet for the roles filter, using both datasets and falling back to the roles in course access if tool usage doesn't load. This isn't ideal because the roles filter will be missing some roles. But maybe it's okay for this edge case.
> 2. make a new result set server-side that just returns the roles - similar to the way the org unit filter has its own result set. This is a bit more work, but will likely behave best and seems more correct. It also has the advantage that the role filter will get populated as early as possible instead of waiting for tool usage data.
> 
> So I'd suggest merging this then working on solution (2). Since time is a bit tight, I could put together a PR for the server-side change if you like.

This new resultset will also help roles legends remain consistent if tools usage runs into intermittent errors. Wonder if that change should also be part of Saad's new PR or a separate one.

PR #242, opened 2022-05-20
Screen Reader Improvments

NicholasHallman:
Functional

Manual
 - Loaded in a page with empty url state, no message
 - Read out the active drill level if it's not the root
 - Don't have aria-loading-progress read out the drill level when the user changes the drill level
 
Integration
 - [x] Check that aria loading has the proper drill message
 
UX
 - Improved screen reader experience
 - Helps screen reader users create a mental model of the dashboard
 
Docs / Security N/A


components/aria-loading-progress.js
```diff
@@ -1,11 +1,11 @@
...
 import { Localizer } from '../locales/localizer';
-import { MobxLitElement } from '@adobe/lit-mobx';
 
-class AriaLoadingProgress extends Localizer(MobxLitElement) {
+class AriaLoadingProgress extends Localizer(LitElement) {
 	static get properties() {
 		return {
-			data: { type: Object, attribute: false }
+			data: { type: Object, attribute: false },
+			isLoading: { type: Boolean, attribute: 'is-loading' }
```
NicholasHallman, line 8: Removed mobx and changed this to observe an is loading so that the component only updates when the data loading state changes isntead of anytime a filter is updated.

bbabu01:
Should org unit drill state be also announced, if applicable, on loading?

components/aria-loading-progress.js
```diff
@@ -24,14 +24,65 @@ class AriaLoadingProgress extends Localizer(MobxLitElement) {
...
 		// loading-start-message is rendered as text node that forces FF to read it, otherwise FF skips it
 		// when loading is finished the code replaces `div` with text node to `div` with aria-label to
 		// to hide loading-finish-message from screen reader (from navigation with arrow keys)
-		const element = this.data.isLoading
-			? html`<div >${this.localize('alert:loadingStarted')}</div>`
-			: html`<div aria-label="${this.localize('alert:loadingDateRange')}"></div>`;
+		const element = this.isLoading
+			? this.renderLoadingMessage()
+			: this.renderLoadedMessage();
+		if (this.isLoading === false) this.hasFinishedLoadingOnce = true;
```
NicholasHallman, line 85: This condition keeps us from notifying the SR about the drill on all sebsequent loads since the toast will handle it.

model/ouDimensionFilter.js
```diff
@@ -53,6 +53,15 @@ export class ouDimensionFilter {
 		return this._selectedOrgUnitIdsAtLevel.get(this._data.parentId) ?? new Set();
 	}
 
+	get selectedItemNames() {
+		const selected = Array.from(this.selectedOrgUnitIds);
+		return this._orgUnitResultSet._data.orgUnits.filter(ou => selected.includes(ou.Id)).map(ou => ou.Name);
```
Vitalii-Misechko, line 58: `this.selectedOrgUnitIds` is set. So, you can write `filter(ou => this.selectedOrgUnitIds.has(ou.Id))`. This way you `selectedItemNames` has O(n) time complexity instead of `O(n*n)`

Vitalii-Misechko, line 58: Why to access private field `_data` when there is a public getter `data`

model/data.js
```diff
@@ -118,6 +118,10 @@ export class Data {
 		return this._orgUnitDrillPath;
 	}
 
+	get isLoading() {
+		return this._resultSets.find(resultSet => resultSet.isLoading === true) !== undefined;
```
Vitalii-Misechko, line 122: This should work the same way
```JS
this._resultSets.find(resultSet => resultSet.isLoading) !== undefined;
```

model/ouDimensionFilter.js
NicholasHallman: This line is mapping the selected id's to the names of the org units in the data, I think that replacing `includes` with `has` would have the same runtime right?

Vitalii-Misechko: Yeah, for this case there is no difference :) in performance. we can avoid unnecessary transformation plus save a few tacts of CPU :) on algorithm

Vitalii-Misechko:
here is what I received when I loaded the dashboard first time (demo page). Case, when there are a few children are selected
URL: http://localhost:8000/demo/?o=3%2C7#
image

Vitalii-Misechko:
Should we construct filter message for screen read ordered by tab order (Org unit filter first, Role filter second, Tool filter third)?

image

johngwilkinson:
> Should we construct filter message for screen read ordered by tab order (Org unit filter first, Role filter second, Tool filter third)?
> 
> image

Seeing this in my email, I had the thought: is reading out the list of org units a bad thing? What if the user has 75 of them selected? I guess the same could be said for roles, although it's less likely. Is there a cutoff where instead of reading them all it will just say that N are selected? Has design had input on this?

NicholasHallman:
> > Should we construct filter message for screen read ordered by tab order (Org unit filter first, Role filter second, Tool filter third)?
> > image
> 
> Seeing this in my email, I had the thought: is reading out the list of org units a bad thing? What if the user has 75 of them selected? I guess the same could be said for roles, although it's less likely. Is there a cutoff where instead of reading them all it will just say that N are selected? Has design had input on this?

No input from design yet, The use case here is when you are given a url with pre-selected filters we should let the user know what filters are there so they have context. I also think we should limit how many get read out to a reasonable limit. I'll run it by design to see what they think that limit should be

Vitalii-Misechko:
Awesome improvements! Like them!

* Screen reader reads filters states when the dashboard loaded
  * checked when each filter is applied individually
  * checked when mixed filters are applied
* The dashboard does not announce that it loads data for the first time.
  * Should we fix that?
* Changes to the filters do not cause that initial message to appear

NicholasHallman:
Okay talked it over with Jocelyne and moved the filter messages from on-load to the descriptions of the buttons. Now if you tab over the role or tool filter it will tell you which filters are selected.

pages/adoption-dashboard-view.js
```diff
@@ -194,6 +194,34 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		if (this.data.isLoading) return;
+		const roleDescription = this.createFilterMessage({
+			filter: this.data.roleFilter,
+			name: this.localize('roleFilter:name'),
+		});
+		const toolDescription = this.createFilterMessage({
+			filter: this.data.toolFilter,
+			name: this.localize('toolFilter:name'),
+		});
+		const roleFilter = this.shadowRoot.querySelector('div.d2l-insights-sections-layout > div.d2l-insights-filter-layouts > div.d2l-dimension-filter-group > d2l-filter:nth-child(2)')?.shadowRoot.querySelector('d2l-dropdown-button-subtle');
```
Vitalii-Misechko, line 211: `lit2` Could you check for null ` this.shadowRoot`?
```JS
setFiltersDescriptions() {
  if (!this.shadowRoot) return;
```

Vitalii-Misechko, line 211: `div.d2l-insights-sections-layout > div.d2l-insights-filter-layouts > div.d2l-dimension-filter-group > d2l-filter:nth-child(2)`

I think big selectors like this are fragile. Someone, who introduces/removes a `div` to improve/fix layout will break screen reader work.
I would suggest introducing a unique class for each filter.

pages/adoption-dashboard-view.js
```diff
@@ -194,6 +194,34 @@ class AdoptionDashboardView extends SkeletonMixin(Localizer(MobxLitElement)) {
...
+		const toolFilter = this.shadowRoot.querySelector('div.d2l-insights-sections-layout > div.d2l-insights-filter-layouts > div.d2l-dimension-filter-group > d2l-filter:nth-child(3)')?.shadowRoot.querySelector('d2l-dropdown-button-subtle');
```
Vitalii-Misechko, line 212: `this.shadowRoot.querySelector(...)?.shadowRoot`
We access a private part of the component therefore, the solution can be easily broken.
I think we should ask Gaudi (or create a PR) to add the desired property to the component so we don't access it in a hacky way

pages/adoption-dashboard-view.js
NicholasHallman: I'm targeting a subtle dropdown inside of the filter component so I can't assign a class to it. I'm trying to override the description property found here https://github.com/BrightspaceUI/core/blob/main/components/filter/filter.js#L224.

I agree it's fragile but I have tests coming that make sure the components are where we expect them to be so if they make changes the tests will fail.

pages/adoption-dashboard-view.js
NicholasHallman: Okay I'll run it by them :)

NicholasHallman:
Removing the button filter summary, will look at this again for M4

test/components/aria-loading-progress.test.js
```diff
@@ -0,0 +1,23 @@
+import '../../components/aria-loading-progress';
+import { aTimeout, expect, fixture, html } from '@open-wc/testing';
+import { Data } from '../../model/data';
+import sinon from 'sinon/pkg/sinon-esm.js';
+
+describe('aria-loading-progress', () => {
+	it('should have a loaded message', async() => {
+		const data = sinon.createStubInstance(Data);
```
Vitalii-Misechko, line 8: TIL

model/data.js
```diff
@@ -118,6 +118,10 @@ export class Data {
 		return this._orgUnitDrillPath;
 	}
 
+	get isLoading() {
+		return this._resultSets.find(resultSet => resultSet.isLoading) !== undefined;
```
Vitalii-Misechko, line 122: `nitpicking` 
```JS
this._resultSets.some(resultSet => resultSet.isLoading)
```

test/visual-diff/adoption-dashboard.visual-diff.js
```diff
@@ -31,12 +31,12 @@ const { snapshot } = require('./diffutils');
...
 			].forEach(errorCommand => {
 				it(errorCommand, async function() {
-					await new Promise(res => setTimeout(res, 1000));
+					await new Promise(res => setTimeout(res, 1300));
 					const rect = await diff.visualDiff.getRect(diff.page, 'body');
 					await diff.page.evaluate((errorCommand) => {
 						window.errorStateTest[errorCommand]();
 					}, errorCommand);
-					await new Promise(res => setTimeout(res, 2000));
+					await new Promise(res => setTimeout(res, 2500));
```
Vitalii-Misechko, line 39: We probably should consider writing a logic that waits on some elements of the page to be drawn instead of timeouts. I see that this test is flaky. No fix is required in this PR.

Vitalii-Misechko (dismissed):
E2E Tests (NVDA)

* Screen reader does not read any messages related to the applied filter until a user focuses the filter.
* Screen reader reads out the drill level except for the root 


Out of scope. Curious if a screen reader should read exact dates in addition to the message `Displaying data between dates`

image

test/visual-diff/adoption-dashboard.visual-diff.js
NicholasHallman: The component we are waiting for here is the highcharts legend. If we add a check for it and a new component gets added that takes longer to update the test will just break anyway so I don't think there's a good solution here

Vitalii-Misechko (approved):

PR #246, opened 2022-05-26
US139406 Show alert if truncation happened on the server side

Vitalii-Misechko:
[US139406](https://rally1.rallydev.com/#/?detail=/userstory/637960043137&fdp=true): [Adoption] Truncate and Notify the user of truncation

Related PR: https://github.com/Brightspace/brightspace-data-platform/pull/106

E2E Tests

* Verify that there is no alert when `isTruncated === false`.
* Modify `bdp-query` that it always return `isTruncated === true`. Verify that alert appears
* Modify the dashboard that it allows to select > 100 org unit. Select > 100 OUs. Verify that alert appears


bbabu01 (approved):
LGTM. Tested with 	isTruncated : true/false in server.js locally

